PR #3, opened 2021-06-06
added color to icons for home page

brittneypostma:


benmccann:
@brittneypostma where are these icons from? I think it'd be nice to add icons for the podcast and newsletter (https://github.com/svelte-society/sveltesociety.dev/issues/242) matching the style that these ones are in

brittneypostma:
I took the icons @kevmodrome had on the site and just added color in Figma. I would be happy to do the other ones as well.

benmccann:
Ok, thanks! I guess it's a question for @kevmodrome then of where did the original logos come from?

PR #4, opened 2021-06-11
Add more pages

benmccann:
These pages are close. The one thing left to do is to fix the layout so that it doesn't take the entire width. But that looks like it's an issue on other pages like /events/2021-summit-spring and should probably be fixed in the common layout separate from this PR


benmccann:
FYI @HAUDRAUFHAUN, I think after this the main thing left to do is the recipes page

swyxio:
hey gang I've added in a placeholder recipes page and fixed up the design of a few of the pages!

here's a quick video with the changes i did

https://www.loom.com/share/c633a5dda5f94e6799e9e3eef77e1b40

pnpm-lock.yaml
```diff
@@ -1,31 +1,7 @@
-lockfileVersion: 5.3
```
benmccann: @sw-yx you could probably drop this file from the commit. It looks like you might be using an older version of pnpm because it downgraded to `lockfileVersion: 5.2`

benmccann (commented):
awesome, thanks! it looks much nicer now

pnpm-lock.yaml
swyxio: ah sorry about that, didn't realize

swyxio:
so do i merge? also what is the gameplan for deploying this thing? do we check in with @kevmodrome ?

benmccann:
I think it'd be fine to merge soon. There's still more work to do, but probably easier for everyone to keep in sync with the changes checked in. I just tried running this locally and did notice one regression from your changes though. The top menu seems to be a bit squished. The events page also doesn't seem to be quite as well aligned as it was before

Screenshot from 2021-06-14 10-36-48

Hopefully we can fix up the recipes page before deploying. That one was a bit trickier. I think it was using some Routify helper that we need to recreate

diepoe:
This might be a problem with bigger screens. I could try fixing it in another pr if you want... :)

benmccann:
I added the recipes in a halfway working state. I added a TODO in the two pages below:
* src/lib/components/recipes/CategoryTree.svelte
* src/routes/recipes/recipes.ts

If anyone would like to take it over and get it over the finish line I won't object :smile:  I think it could be helpful to merge this PR eventhough it's not quite done. It'd probably make it easier to collaborate and send additional PRs against the current state of things

swyxio:
@ben i have fixed the css for the events page and made some tweaks to the recipes page to make it a little nicer

<details>
<summary>screenshots
</summary>


CleanShot 2021-06-15 at 06 28 45

CleanShot 2021-06-15 at 06 29 14



</details>

benmccann:
Cool. That looks much nicer! I finished populating the recipes page, so I'm afraid it looks a bit off again with the new rest of the data if you want to take another look at the CSS

There's one more thing left to do, which is that the table of contents on the recipes pages are missing. It looks to me like that's contained in `src/lib/layouts/Recipe.svelte` and `src/lib/layouts/RecipeCategory.svelte` and is maybe set in the `mdsvex` config in `svelte.config.js`. I've never used mdsvex before, so I'm not quite sure how that gets wired up

swyxio:
put some circles on the icons and squared them up.. a little trick i learned from steve schoger 


<details>
<summary>screenshot
</summary>

CleanShot 2021-06-15 at 07 11 01

</details>

benmccann:
Nice! Much better :smile: 

I just pushed the recipe table of contents. Some of the recipes are showing up with horizontal scrollbars. Also, the icons are missing from the table of contents because they're trying to load as `img` instead of `svg`, but that's also broken on the current site, so shouldn't be a blocker

swyxio:
im fine merging... will do after one more UI/UX review later

benmccann:
I'm fine merging as well

swyxio:
great work team!!

PR #7, opened 2021-06-17
Remove duplicate link to podcast

benmccann:
We now have a big icon for the podcast just above, so no longer need to separately mention that we "also run the Svelte Radio podcast".


benmccann:
Closing this. It makes more sense when you read the text (which I never do :smile:)  The text duplicates the menus and icons quite a bit and if we were going to change it then it'd need a bigger change than just this

PR #8, opened 2021-06-17
Cleanup component categories

benmccann:
Ref https://github.com/svelte-society/sveltesociety.dev/issues/218

There was an "Integration" and "Integrations" categories. I'd like to make one "SvelteKit Adapters", which is more specific and will allow us to link to https://sveltesociety.dev/ from the SvelteKit docs

There was only one in "component sets", which I think was accidentally putting a tag in a category

There were two "Unclassified" sections. One generated by an empty string and one where "Unclassified" was specifically written

### DIFF
--- a/src/routes/components/components.json
+++ b/src/routes/components/components.json
@@ -9,7 +9,7 @@
       "cli tools"
     ],
     "addedOn": "2021-03-31T00:00:00Z",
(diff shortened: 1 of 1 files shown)


kevmodrome:
Thanks Ben, greatly appreciated!

PR #11, opened 2021-06-24
Add SvelteKit boilerplates and adders

benmccann:
I also removed the "templates" tag. It wasn't very informative since all boilerplates are templates


kevmodrome (approved):
LGTM

PR #26, opened 2021-07-25
fix: updated github repository link

abosch19:
changed links to new sveltesociety-2021 github project

### DIFF
--- a/src/routes/help/components.svx
+++ b/src/routes/help/components.svx
@@ -8,15 +8,15 @@ h2 { font-size: 23px; }
 
 To add a new component on the website, the process is rather simple.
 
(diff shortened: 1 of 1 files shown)


benmccann:
I feel like `sveltesociety-2021` will age badly as a repository name and that we should rename the current `sveltesociety.dev` repo to something like `sveltesociety-legacy` and then rename this one to `sveltesociety.dev`, but that would have to be up to @kevmodrome

kevmodrome:
I think that makes sense @benmccann 👍

kevmodrome:
Sorted out. Thanks for the PR though! 🙏

PR #28, opened 2021-07-25
Added Event Calendar (@event-calendar/core)

vkurko:


kevmodrome (changes requested):
Please don't add new cateogories or tags, make it work with what is already there 🙏

vkurko:
> Please don't add new cateogories or tags, make it work with what is already there 🙏

I checked the category and tags and it seems that there were only 2 extra tags that I already removed.

kevmodrome (approved):

Issue #5, opened 2021-06-17
Parity with sveltesociety.dev


benmccann:
Here's a list of things I noticed that are different between this repo and the production site:

* ~~favicon is missing (https://github.com/svelte-society/sveltesociety-2021/pull/13)~~
* ~~page titles are missing - for homepage, components, boilerplate, cheat sheet, events (https://github.com/svelte-society/sveltesociety-2021/pull/13)~~
* ~~`robots.txt` - renders the homepage instead of a valid robots.txt file (https://github.com/svelte-society/sveltesociety-2021/pull/19)~~
* ~~events page - most past events are missing (https://github.com/svelte-society/sveltesociety.dev/pull/36)~~
* ~~recipes - text is very large on individual recipe pages and recipe category pages (https://github.com/svelte-society/sveltesociety.dev/pull/32)~~
* boilerplates and components - controls are misaligned and missing arrow icon ~~(https://github.com/svelte-society/sveltesociety.dev/pull/41)~~
* ~~CSS stacking issue? If you visit all the pages a couple times you will start to see the page squished to the left hand side of the screen. Is there some CSS from an old page not being removed? (https://github.com/svelte-society/sveltesociety.dev/pull/42)~~
* ~~`npm run build` fails (https://github.com/svelte-society/sveltesociety.dev/issues/35)~~

kevmodrome:
This is a good list. I'll post on Twitter and on Discord and see if we can get some people interested in helping with this.

christianp86:
Happy to help 🙌🏻

benmccann:
That'd be great @christianp86! Any idea when you might be able to take a look?

christianp86:
@benmccann I will take a look tonight. Should we create individual issues for better tracking?

benmccann:
no need. I can just update the list above as PRs are merged to cross out items that are done

christianp86:
I'll try fixing some more issues this week 👍🏻

christianp86:
Sorry guys but finding time during my parental leave is harder then expected 🙃 I did some additional fixes for events but then my little girl woke up from her nap 😴 
Long story short I am on vacation for two weeks now and will get back to it once we are home again. Cheers chris

benmccann:
@kevmodrome there's a good handful of PRs pending. Once they're all merged I think we should be good to launch!!

kevmodrome:
Closing this since we've launched!

PR #41, opened 2021-08-02
Fix: boilerplates and components

ambarvm:
Fix item from #5: boilerplates and components - controls are misaligned and missing arrow icon


benmccann:
Wow, you're the best! I looked to see what else might be missing and it looks like there's also a `search-icon.svg` that is referenced in `boilerplates/index.svelte` and `components/index.svelte` that should be copied over

benmccann (approved):

PR #42, opened 2021-08-02
Fix: page squished to left after opening tooling page

ambarvm:
Fix item from #5

Global styles were affecting all components once tooling was opened and `max-width: 65ch;` was being applied on each page.


benmccann (approved):
fantastic!

kevmodrome (approved):

Issue #51, opened 2021-08-06
Styling updates


brittneypostma:
Originally css variables were used to style inside the style tags with vanilla CSS. However, tailwind has been added and most files need to be updated. There are lots of optimizations that could happen as well by adding more top level styles to tailwind config instead of styling individual elements.

brittneypostma:
merged

PR #53, opened 2021-08-07
Styles

brittneypostma:
Updated tailwind config to include css variable styles that might be used in the future, converted header, layout, index, and the cheat sheet page to what was discussed in society-chat in discord. I tried moving some of the styles higher up to make styling more consistent, but didn't change anything on the routes other than index and cheat sheet.

This will be a personal choice if you want to move more towards using tailwind. I put the variables into tw config that could be used to convert other files over and created a color palette based on what you used in css-variables before.

Issue #51

### DIFF
--- a/src/app.postcss
+++ b/src/app.postcss
@@ -6,15 +6,18 @@
 @font-face {
 	font-family: Inter;
 	src: url(/fonts/Inter-Regular.woff2);
+	font-display: swap;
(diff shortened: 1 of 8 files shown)


kevmodrome:
Personally not a fan at all of TW but it seems like a bunch of people like it so I'm open to it.

PR #56, opened 2021-08-09
Use component page pattern for tooling

### LINKED ISSUE #45: Should 'tools' follow the pattern of components and templates
The tools section follows a different UI pattern, it would be nice to use the same pattern

ambarvm:
Fixes #45


kevmodrome:
Thanks

PR #75, opened 2021-08-24
upgrade dependencies

benmccann:
Upgrade dependencies to pull in a new version of `@sveltejs/adapter-static`. This will make it so that output is cleared between builds, which will help @userquin's work on better supporting service workers


kevmodrome:
Excellent! Didn't realise this had been solved, I've run into issues with the adapter before that I think is related to this.

PR #80, opened 2021-08-29
Add svelte-number-spinner

bohnacker:


kevmodrome:
Thanks for the PR but it does need some fixing, change the NPM value to just the package name! 👍

bohnacker:
Thanks for pointing that out! 👍

PR #90, opened 2021-08-30
refactor: centralize sorting logic

thollander:
- Sorting logic was repeated/spreaded across multiple components
- Extract the logic from Svelte components 
- Unit test it thanks to jest

### DIFF
--- a/src/routes/components/index.svelte
+++ b/src/routes/components/index.svelte
@@ -11,7 +11,7 @@
   const allCategories = Array.from(new Set(components.map(item => item.category).flat()))
   let filterTag = []
   let filterCategory = null
-  let sorting = 'added_desc';
+  let sorting = 'stars_desc';
   let packageManager = 'npm'
 
   const intersection = (array1, array2) => {
--- a/src/routes/components/index.svelte
+++ b/src/routes/components/index.svelte
@@ -3,6 +3,7 @@
   import List from "$lib/components/ComponentIndex/CardList.svelte";
   import Button from "$lib/components/ComponentIndex/ArrowButton.svelte";
   import components from "./components.json";
+	import { compare } from '$lib/utils/sort';
 
   let searchValue;
   let searchTag;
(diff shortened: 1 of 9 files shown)


src/routes/components/index.svelte
```diff
@@ -3,6 +3,7 @@
   import List from "$lib/components/ComponentIndex/CardList.svelte";
   import Button from "$lib/components/ComponentIndex/ArrowButton.svelte";
   import components from "./components.json";
+	import { compare } from '$lib/utils/sort';
```
benmccann: it looks like this file is indented using spaces, but this line is a tab

kevmodrome:
Thanks, this looks great! 

Though I do not think we should use Jest. In the spirit of Svelte let's try to use something like uvu instead: https://github.com/lukeed/uvu

benmccann:
I'm not totally against Jest personally. Some of the other core Svelte maintainers don't like it, but I think it has a much bigger community, is better documented, etc. E.g. you can use `svelte-jester` to handle files that need pre-processing. I have no idea how you'd do that with uvu since there are no examples of it.

kevmodrome:
I suppose you're right @benmccann - lets go with Jest. It'll make it easier to get contributions.

kevmodrome:
I'll go ahead and merge this. Can you sort out the conflicts when you have time @thollander ? Love the work done here.

ambarvm:
uvu has a [svelte example](https://github.com/lukeed/uvu/tree/master/examples/svelte) in their repo

benmccann:
> uvu has a svelte example in their repo

Yeah, but it doesn't handle preprocessing. E.g. if you want to write TypeScript in your template I'm not sure how you'd do that with uvu. I'm sure you can, but there's just way less documentation

benmccann:
Btw, I think it'd be really cool to make the default sorting by number of stars. Right now I think it's by date added to sveltesociety.dev, but that's kind of random looking and not really meaningful to users

kevmodrome:
Default sort by stars does sound better, agreed! 

For some reason I thought it was sorted by date published, silly me 😆

src/routes/components/index.svelte
thollander: thx, done ✅

benmccann (approved):

PR #93, opened 2021-08-30
fix: no stars print if undefined

thollander:
## Bug : 
1. go to https://sveltesociety.dev/components/
2. Sort by "Stars Asc"
3. See “undefined" as stars number

image

The issue is there because the stars are secured to `0` only on the initialization : https://github.com/thollander/sveltesociety.dev/blob/0aee085e7428ca5ae60e6a5f0d3224779b3c7bc2/src/lib/components/ComponentIndex/Card.svelte#L7

## Fix : 
To fix this issue, I totally didn't print the section if we don't know the number of stars (not relevant to set 0, if we actually don't know).

### DIFF
--- a/src/lib/components/ComponentIndex/Card.svelte
+++ b/src/lib/components/ComponentIndex/Card.svelte
@@ -95,7 +95,7 @@
       <Tag title={tag} variant='blue' />
     {/each}
   </div>
-  {#if stars > 0}
+  {#if typeof stars !== 'undefined'}
     <div class="card__bottom">
       <div>
         {#if (repo || url).includes('github')}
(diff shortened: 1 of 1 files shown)


src/lib/components/ComponentIndex/Card.svelte
```diff
@@ -95,22 +95,22 @@
       <Tag title={tag} variant='blue' />
     {/each}
   </div>
-  <div class="card__bottom">
-    <div>
-      {#if stars > 0}
+  {#if stars > 0}
```
benmccann: 0 is different than `undefined`. can we still show it if `0` but no `undefined`? maybe change to `{#if typeof stars !== 'undefined'}

thollander: Well spotted ! 
Tested it locally, seems fine : 
image

benmccann (approved):

PR #94, opened 2021-08-31
Feature/design system

brittneypostma:
Look this over for changes. Left old css file in lib/styles/old-app.css to review. New styles are root, globals, and reset css files. Updated most of the tailwind stuff out, but I may have missed some and replaced old css variables with updated ones.


benmccann:
I preferred the wider screen width and smaller font in `main`. Things look a bit scrunched with the PR. E.g. one of the arrows in the controls isn't displaying because there's not enough room.

**`main`**

Screenshot from 2021-08-31 08-31-48

**PR**

Screenshot from 2021-08-31 08-31-36

src/lib/components/EventListElement/index.svelte
```diff
@@ -29,20 +29,26 @@
...
 	.event-tile {
 		--tw-shadow: 0 4px 6px -1px rgba(0, 0, 0, 0.1), 0 2px 4px -1px rgba(0, 0, 0, 0.06);
 		box-shadow: var(--tw-ring-offset-shadow, 0 0 #0000), var(--tw-ring-shadow, 0 0 #0000),
 			var(--tw-shadow);
 		padding: 2rem;
 		margin-top: 3rem;
 		border-radius: 0.5rem;
-		background: #f3f6f9;
+		background: var(--white);
+    display: grid;
```
benmccann, line 43: mix of tabs and spaces added throughout this PR

brittneypostma: I can definitely update the max-width and may have missed some of the changes on individual pages. I went through them, but quiet a few need updated to the new variables.

brittneypostma: I think all of this can be adjusted to our liking after we settle on the design system and get it implemented. There are lots of little tweaks that will need to be made.
I do suggest we update the box-shadow naming too, they are confusing to me. I would prefer to go with a sizing scale for them and then we have a random one created in the EventListElement.

benmccann: I think if we can at least fix the max-width and font size in the PR that should be good for now and we can do the rest in follow ups

PR #95, opened 2021-08-31
Feature/design system

### LINKED ISSUE #91: Remove Tailwind and set up a design system using css custom properties
This is a quite a big one but it needs to be done to make it easier in the future.

- [ ] Add proper CSS custom properties for regular use (colors, grays, sizing, elevation, shadows)
- [ ] Define common styles for headings (h1, h2, h3, etc.) and other common elements
- [ ] Remove Tailwind (including all the preprocessing and files)

Edit: changed title to not be so rude.

brittneypostma:
@kevmodrome asked for a redo of pr so the preview would work. 
[Old PR](https://github.com/svelte-society/sveltesociety.dev/pull/94)

Fixes #91


benmccann:
Copying my comments over from the old PR... I preferred the wider screen width and smaller font in `main`. Things look a bit scrunched with the PR. E.g. one of the arrows in the controls isn't displaying because there's not enough room.

**`main`**

Screenshot from 2021-08-31 08-31-48

**PR**

Screenshot from 2021-08-31 08-31-36

src/lib/components/EventListElement/index.svelte
```diff
@@ -29,20 +29,26 @@
...
 	.event-tile {
 		--tw-shadow: 0 4px 6px -1px rgba(0, 0, 0, 0.1), 0 2px 4px -1px rgba(0, 0, 0, 0.06);
 		box-shadow: var(--tw-ring-offset-shadow, 0 0 #0000), var(--tw-ring-shadow, 0 0 #0000),
 			var(--tw-shadow);
 		padding: 2rem;
 		margin-top: 3rem;
 		border-radius: 0.5rem;
-		background: #f3f6f9;
+		background: var(--white);
+    display: grid;
```
benmccann, line 43: it looks like this PR has a mix of tabs and spaces throughout

kevmodrome:
Looks good @brittneypostma 

I don't now if we want to have so many spacing vars, we can probably reduce them by half. The other stuff is 👌 I'm with you on the shadow vars, feel free to change them to whatever you prefer! Change them to some other styles as well if you want.

brittneypostma:
I used the tailwind spacing scale so that it is easier for those of us familiar with tw.  Probably could fine tune it to what we want to use.

brittneypostma:
Don't push buttons on mobile 😂 sorry

kevmodrome:
When it comes to the font sizes I think we want the opposite of what it is now. They should grow when the screen size gets smaller.

If we wanted to we could also have different space values for big and small screens as well. We would probably need something fixed as well though. So (just to illustrate):

Large screens:
--s-xs: 4px;

Small:
--s-xs: 8px;


And then static (f = fixed):
--s-xs-f: 4px;

benmccann:
I can't think of any site that makes the font bigger when the screen is smaller. Can you name an example where that happens? It sounds quite funny to me. I'd expect the font size to remain the same

brittneypostma:
I used the font sizes that were already declared. They use clamp and are smaller on large screens and larger on large screens. This is typical for font sizes. 

The sizing I'm less sure what you mean. I have never used clamp for that since growing isn't always what is expected with paddings and margins. Adding a fixed variable would add even more properties to the list we want to reduce. I usually update these in a media query where it is needed.

brittneypostma:
@benmccann I was thinking the same. The font size usually gets bigger as the screen gets larger to help readability, either through clamp or a media query.

kevmodrome:
> I used the font sizes that were already declared. They use clamp and are smaller on large screens and larger on large screens. This is typical for font sizes.
> 

Yep! I think I added them way back at some point.

> The sizing I'm less sure what you mean. I have never used clamp for that since growing isn't always what is expected with paddings and margins. Adding a fixed variable would add even more properties to the list we want to reduce. I usually update these in a media query where it is needed.

> I can't think of any site that makes the font bigger when the screen is smaller. Can you name an example where that happens? It sounds quite funny to me. I'd expect the font size to remain the same

> @benmccann I was thinking the same. The font size usually gets bigger as the screen gets larger to help readability, either through clamp or a media query.

I don't know how common it is, honestly. I've just been deep in the weeds of Adobe's Spectrum design system lately and they do this: https://spectrum.adobe.com/page/typography/#Font-sizes

We don't need to use clamp for it either. Anyway, I'm fine with whatever works well in our particular case. If you think smaller fonts on mobile works better, then go with that! 

> The sizing I'm less sure what you mean. I have never used clamp for that since growing isn't always what is expected with paddings and margins. Adding a fixed variable would add even more properties to the list we want to reduce. I usually update these in a media query where it is needed.

Yeah, I should probably clarify a bit here. I don't mean that we should use clamp for the spacing (though we probably could if we wanted to, when I've tested it it actually works out quite well). We want some differences in sizing depending on the users screen. For example a button needs more padding on mobile than on desktop to make sure that the clickable area is large enough. So what I propose is something like this:

```css
:root {
    --s-xs: 6px;
    ...

    --s-xs-f: 4px;
    ...
    --s-xxl-f: 48px;
}

@media big screen {
    :root {
        --s-xs: 4px;
    }
}
```

In cases where we want to make sure users can click things we point to the dynamic scale, that gets bigger on mobile. I don't think we necessarily need more than 6 or 7 sizes. Remember that we want to restrict the ability of the contributors to a smaller set of options that makes it easier to contribute as well as maintain.

Edit: Also I don't necessarily think we need to adress all these in this PR, we can do that later as well.

brittneypostma:
I'm updating the max-width now and removing the spacing variables with odd decimals as I don't think we will need them. Other than that, I think we should figure out what scale is going to be used to keep consistency and cut out the ones outside of that scale.

benmccann:
I like that the top menu extends across the screen now. I also like that it's dark blue now. I think the content on the homepage could be constrained in width a bit like the current homepage

The fonts still feel way too big.

brittneypostma:
> I like that the top menu extends across the screen now. I also like that it's dark blue now. I think the content on the homepage could be constrained in width a bit like the current homepage
> 
> The fonts still feel way too big.

I agree on the fonts and need to sit down with it today and actually go through it all. I will get another commit in for that today, it was just more than I could get to last night. 

I will also constrain the home page to the original size and leave the other ones larger.

brittneypostma:
I ran `npm run format` because Ben mentioned having a mix of tabs and spaces. It made 36 modifications. I'm not going to put this into the pr, but maybe we can be run after to fix all these issues? We could add something like husky to do auto formatting when pushing.

I updated the font-sizes and made them a bit smaller, I think it fits more naturally now. I also fixed a few things that weren't showing correctly and adjusted the max-width. Let me know if there is anything else that needs changed.

benmccann (approved):
nice! it looks good to me! thanks for all the help with this

brittneypostma:
@kevmodrome are we good on this or does anything else need to be changed in this pr?

kevmodrome:
I'm fine with merging this now if you are @brittneypostma

brittneypostma:
> I'm fine with merging this now if you are @brittneypostma

Sounds good to me, I might need to resolve some conflicts from these other prs.

brittneypostma:
Just pushed the last tweaks to this from the last pr, have a look at the hover menus and see if the styling looks decent. I spoke with @kevmodrome about creating a new issue to overhaul that entire section. If you two are happy with where it site, I say we merge this and I'll work on the little tweaks and updates after.

brittneypostma:
All set to merge now, I believe. 😀

PR #101, opened 2021-09-06
styling tweaks around the select component, created a new issue regar…

brittneypostma:
styling tweaks around the select component, created a new issue regarding the difficulties with this package

### DIFF
--- a/src/lib/components/ComponentIndex/ArrowButton.svelte
+++ b/src/lib/components/ComponentIndex/ArrowButton.svelte
@@ -6,6 +6,7 @@
 
 <style>
   div {
+    align-self: flex-end;
     font-size: 1rem;
     border: 2px solid var(--dark-gray);
     border-radius: 3px;
@@ -15,7 +16,7 @@
     display: flex;
     align-items: center;
     justify-content: space-between;
-    padding: var(--s-2);
+    padding: 13.5px var(--s-2);
     position: relative;
   }
   div.small {
--- a/src/lib/components/ComponentIndex/ArrowButton.svelte
+++ b/src/lib/components/ComponentIndex/ArrowButton.svelte
@@ -6,11 +6,9 @@
 
 <style>
   div {
-    align-self: flex-end;
     font-size: 1rem;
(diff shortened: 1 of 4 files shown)


benmccann:
I like that these are more like regular select components now. They're much harder to use though because the text is being cut off after eight characters and continued with ellipses. Is there a way to make the dropdown extend horizontally past the select component so that we can see more of the text. Also, it'd be nice to switch the package manager one as well so that they're all consistent.

brittneypostma:
> 
> 
> I like that these are more like regular select components now. They're much harder to use though because the text is being cut off after eight characters and continued with ellipses. Is there a way to make the dropdown extend horizontally past the select component so that we can see more of the text. Also, it'd be nice to switch the package manager one as well so that they're all consistent.

This is exactly what I was referring to, I created an issue to not use that Svelte Select package. It is difficult to style consistently and it doesn't seem to have uncomplicated anything. This is just minor changes to at least make the buttons look better until we remove it.

benmccann:
I wonder if we should just wait until then? I think this PR makes it look better in some ways, but makes it much harder to actually use due to that limitation

kevmodrome:
Let's add an auto column grid here with a min width to achieve a two per row kind of thing here. That should also make it display nicely on mobile.

brittneypostma:
> 
> 
> Let's add an auto column grid here with a min width to achieve a two per row kind of thing here. That should also make it display nicely on mobile.

So you want the inputs stacked on mobile and then 2 row with 2 inputs on each row on larger screens?

brittneypostma:
I updated this to what I think you meant @kevmodrome and changed the searchbar to be more inline with the style of the inputs.

benmccann:
this looks really nice. I might switch the package manager select as well, but other that this looks really good

kevmodrome:
Sorry, been a bit busy lately (I know, recurring theme) - looks good.

brittneypostma:
Sweet, no worries, I just started this week at my new job so I've been the same. I'm going to merge this, to avoid any conflicts with the other pr's. I can help take a look at those this weekend.

PR #106, opened 2021-09-07
Fetch stars from Gitlab GraphQL API

ambarvm:
Script will now update both github and gitlab stars


brittneypostma:
Does this require a rebuild to show the correct stars? Just checked one off the deploy preview and it wasn't correct. Not sure how the flow for these pages is working, but I would assume we need to be fetching this data in a load function and I don't see that being used.

benmccann:
you have to manually run the script for now. that's not too bad because it keeps us from hitting api limits and the page is prerendered anyway. it's be interesting to look at moving this to `load`, but it sounds like a separate issue

PR #107, opened 2021-09-07
add component svelte-toc

janosh:
https://github.com/janosh/svelte-toc


brittneypostma:
I think this is fine to add if you can resolve the conflict.

janosh:
Should be all set now.

PR #108, opened 2021-09-07
Add component snippet generator

aantusahaa:
Component Snippet Generator

# Why?
Currently, a lot of copy-pasting is necessary to get the snippet ready for the pull request. And there is room for errors as well, especially in the `category` and `tags` property. In `tags` there is an instruction that says, "avoid creating new tags" and in `category` there is an instruction to add only predefined categories. But there is no easy way to find all of the currently available tags except for the `Tags` component on the `components` page. It is not great UX-wise. That's why I made this effort to make the snippet generation process a little bit more user-friendly.


src/routes/help/components.svelte
```diff
@@ -0,0 +1,243 @@
+<script>
+	import components from '../components/components.json';
+
+	const allTags = Array.from(new Set(components.map((item) => item.tags).flat()));
+	const tagItems = allTags.map((t) => ({ label: t, value: t }));
+	const allCategories = [
+		'Boilerplate',
```
benmccann: I don't see any category called "Boilerplate". Where is this coming from?

benmccann (commented):
This looks nice. The main issues I see is that you're also taken to this generator from the templates and tooling pages and they should probably have their own. It would also be nice if the categories were dynamically generated rather than hardcoded

src/routes/help/components.svelte
aantusahaa: It is in the current staging site. This is the reason I didn't make the categories dynamic. There is this inconsistency here, on `/help/components` the list is `Boilerplate, Data Visualisation, Design Pattern, Design System, Developer Experience, Forms & User Input, Integration, Routers, Stores, Testing, User Interaction` but on `/components` the dynamically generated list contains `Forms & User Input, User Interaction, SvelteKit Adapters, Routers, Integration, Developer Experience, Design System, Design Pattern, Stores, Data Visualisation, Testing, Animations, Rich Text Editor`. I'm not sure which one to follow.

image

aantusahaa:
> The main issues I see is that you're also taken to this generator from the templates and tooling pages and they should probably have their own.

I'd love to do so. But need some instructions regarding which fields to put there if they are not the same as components. It'll be nice to do one for `templates` as well I think.

That got me thinking, wouldn't it be nice to have only one page for there of these? We could put one extra field there to select the type of snippet to generate(one of the templates, components, tooling) and based on that selection we could filter the tag and categories list accordingly. Because in current state only these two properties are different across these three types.

src/routes/help/components.svelte
benmccann: `/help/components` is outdated. Please use the list from `/components` instead. That should allow you to make it dynamic as well

aantusahaa: Committed. 
Should I sort categories and tags alphabetically or leave them as is? I assume it'll be easier to scan if it's sorted.

benmccann: sorting makes sense to me

aantusahaa: Committed.

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+		addedOn,
+		category
+	};
+
+	function ascending(a, b) {
+		return a.label > b.label ? 1 : -1;
+	}
+</script>
+
+<h1>How to submit a new component?</h1>
```
benmccann: Some headings are questions and some are not. Let's make none of them questions for consistency
```suggestion
<h1>Submitting a new component</h1>
```

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+</p>
+
+<h2>Forking the project</h2>
+<p>
+	You can fork the <a href="https://github.com/svelte-society/sveltesociety.dev/">
+		GitHub project
+	</a>, add your component and then propose a Pull request.
+</p>
+
+<h2>Edit the file</h2>
```
benmccann: the other headings are actions, so we could make this one as well for consistency
```suggestion
<h2>Editing the file</h2>
```

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+<p>
+	You can edit and propose your changes <a
+		href="https://github.com/svelte-society/sveltesociety.dev/edit/master/src/pages/components/components.json"
+	>
+		directly in GitHub
+	</a>
+</p>
+
+<h2>What information should I give ?</h2>
```
benmccann:
```suggestion
<h2>Generating file contents snippet</h2>
```

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+<p>
+	To add a new component on the website, the process is rather simple. You have to add your
+	component in <a
+		href="https://github.com/svelte-society/sveltesociety.dev/blob/master/src/routes/components/components.json"
+	>
+		components.json
+	</a>
+</p>
+
+<h2>Forking the project</h2>
```
benmccann: I would remove this section. Manually forking the project is not the easiest way to do this. The next section already talks about editing the file directly on Github which is easier and will automatically fork the project for you

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+	</a>
+</p>
+
+<h2>What information should I give ?</h2>
+<p>
+	Each component is represented by a JSON Object. Use the generator below to generate the JSON
+	Object.
+</p>
+
+<h2>Generator</h2>
```
benmccann: We could put the "Each component is represented by a JSON Object. Use the generator below to generate the JSON Object." description and generator under the same heading

```suggestion
```

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+	You can edit and propose your changes <a
+		href="https://github.com/svelte-society/sveltesociety.dev/edit/master/src/pages/components/components.json"
+	>
+		directly in GitHub
+	</a>
+</p>
+
+<h2>What information should I give ?</h2>
+<p>
+	Each component is represented by a JSON Object. Use the generator below to generate the JSON
```
benmccann: Please keep "Object" on this line and don't manually wrap the text. Text wrapping can be done by VS Code or whatever editor is being used to edit this file

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+	</div>
+	<div class="input-wrapper">
+		<label for="adden-on">Added On*:</label>
+		<div>
+			<input id="adden-on" type="date" required bind:value={addedOn} />
+			<span class="input-helper">
+				The date when the component have been added on the website (generally it’s today)
+			</span>
+		</div>
+	</div>
```
MacFJA: I'm not sure that field should be editable

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+<p>
+	Each component is represented by a JSON Object. Use the generator below to generate the JSON
+	Object.
+</p>
+
+<h2>Generator</h2>
+<p><code>*</code> marked fields are required</p>
+<div class="json-generator">
+	<div class="input-wrapper">
+		<label for="title">Title*:</label>
```
MacFJA: Instead of adding `*` to every label of required field, I would do it with CSS:
```css
.required:after {
  content: "*";
  color: red;
}
```
```html
<label class="required" for="title">Title:</label>
```

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+		<label for="desc">Description:</label>
+		<div>
+			<input id="desc" type="url" bind:value={description} />
+			<span class="input-helper">A short description of the component</span>
+		</div>
+	</div>
+	<div class="input-wrapper">
+		<label for="npm">NPM:</label>
+		<div>
+			<input id="npm" type="url" bind:value={npm} />
```
MacFJA: The NPM should NOT be an URL but the package name as it is in NPM/Yarn registry

src/routes/help/components.svelte
```diff
@@ -0,0 +1,238 @@
...
+	let clipboardCopy = false;
+	const copyToClipboard = (text) => {
+		navigator.clipboard
+			.writeText(text)
+			.then(() => {
+				clipboardCopy = true;
+				setTimeout(() => (clipboardCopy = false), 1000);
+			})
+			.catch(() => alert('Clipboard copy Permission denied'));
+	};
```
MacFJA: This function should be put inside a dedicated file as it is used in several file now

src/routes/help/components.svelte
aantusahaa: Removed this section.

src/routes/help/components.svelte
aantusahaa: I'm indeed using VS Code. I don't know what is causing this hard wrap on format. `printWidth` is 100 inside `.prettierrc`, so that hard wrap shouldn't happen. For now, I replaced _JSON Object -> Object_. Still conveys the meaning I think. Please let me know how to fix this If you have any idea.

src/routes/help/components.svelte
aantusahaa: I'm not sure about it either. Currently, the date is not being displayed anywhere. And I think there is either a plan to automate the process or remove this field altogether. Need some more thoughts on this.

src/routes/help/components.svelte
aantusahaa: Thanks for the suggestion. Committed that.

src/routes/help/components.svelte
aantusahaa: Silly me! Fixed now.

src/routes/help/components.svelte
aantusahaa: I agree. Along with this, there are several other logic duplications (allTags, allCategories) across the project. They all should Ideally be put inside a separate file. Maybe subject for a different PR?

MacFJA: I just made it: #115 😉

src/routes/help/components.svelte
benmccann: I don't think it matters either way. The user is going to copy/paste the result and can edit it then, so there's not much point to preventing them from editing it. As long as the default is the current date it's probably fine

src/routes/help/components.svelte
```diff
@@ -0,0 +1,242 @@
...
+	</div>
+	<div class="input-wrapper">
+		<label for="npm">NPM:</label>
+		<div>
+			<input id="npm" type="text" bind:value={npm} />
+			<span class="input-helper">The npm name of the component</span>
+		</div>
+	</div>
+	<div class="input-wrapper">
+		<label for="img">Image:</label>
```
benmccann: We should remove this. There aren't any. The only instances of it were empty. I just removed them: https://github.com/svelte-society/sveltesociety.dev/commit/91bf13080ddec8031ae400f6bb71ef504dac306c

benmccann:
lgtm. I think we just need it to handle templates and tooling

> That got me thinking, wouldn't it be nice to have only one page for there of these? We could put one extra field there to select the type of snippet to generate(one of the templates, components, tooling) and based on that selection we could filter the tag and categories list accordingly. Because in current state only these two properties are different across these three types.

that sounds like a good plan to me

aantusahaa:
Since this page is now handling templates, components and tooling, the route should change too. My suggestion would be `/help/submit`.

benmccann:
How about `/help/submitting`?

aantusahaa:
> How about `/help/submitting`?

Sounds good to me. Should I go ahead with this?

benmccann:
Yeah, I'd say go ahead and rename it

aantusahaa:
Done. Please let me know if I missed something.

benmccann:
Cool. This lgtm. One thing I noticed that might be nice would be if you could click a previously selected item in the tags list to deselect it

aantusahaa:
> One thing I noticed that might be nice would be if you could click a previously selected item in the tags list to deselect it

Maybe no longer necessary. `svelte-select` provides superior UX than the default select. And most importantly cross-browser consistency.

benmccann:
One last thing I just noticed. The sentence "You can edit and propose your changes directly in GitHub" always points at `components.json` even if you've selected Template or Tool. Similarly, there's also a sentence "You have to add your component in components.json"

Maybe we should move this down below the JSON snippet and say something about adding it to the file there instead, but have the file name update appropriately with the dropdown

aantusahaa:
Made those links dynamic. There are still some `component` words left above the generator. Changing those words in relation to the dropdown would be jarring to the eye. It would be better if we can come up with a word that will encompass all those types or change the phrases entirely. Some suggestions would be great.

And in the matter of links can we change `tooling -> tools` for consistency?

aantusahaa:
One more thing I noticed. If you click "Submit a template" it will navigate you to `/help/submitting?type=template` and the dropdown will be set to Template. Same for component and tool. App is getting the query string with `$page.query.get('type')`. Now if you visit the link directly or refresh the page it will become `/help/submitting/?type=template`. Notice the extra forward slash before query string. As a result the function might fail to get the query. It doesn't happen in dev server only in live site. 
Is this expected behaviour?

benmccann:
> Made those links dynamic. There are still some component words left above the generator. Changing those words in relation to the dropdown would be jarring to the eye. It would be better if we can come up with a word that will encompass all those types or change the phrases entirely. Some suggestions would be great.

I'd just leave it as is. I think it reads fine now

> And in the matter of links can we change tooling -> tools for consistency?

That makes sense to me

> It doesn't happen in dev server only in live site. Is this expected behaviour?

Maybe file an issue about this. I wonder if there is a Netlify setting controlling this

aantusahaa:
> Maybe file an issue about this. I wonder if there is a Netlify setting controlling this

Whether there is an extra forward slash or not `$page.query.get()` should be able to pull out the query string right?

benmccann:
> Whether there is an extra forward slash or not $page.query.get() should be able to pull out the query string right?

Looks like that's a bug in SvelteKit. I'll file an issue for it

PR #111, opened 2021-09-09
Persist package manager option in local storage

ambarvm:
Link: https://deploy-preview-111--sveltesociety-preview.netlify.app/components/


benmccann:
just curious, why local storage vs cookie? I'm not that familiar with when you'd use one vs the other

ambarvm:
Cookies are sent to the server with every http request vs Localstorage is javascript only. So localstorage over cookie because this is not of any use to the server.
I guess cookie would be more appropriate if this was server rendered.

benmccann:
It is server-rendered, but will be updated by hydration on the client-side

ambarvm:
I thought we are using a static export. 🤔

If it is server rendered I will try to do this with cookie.

benmccann:
Oh, I guess it is a static export in production actually. I assumed we'd be using `adapter-netlify` since it's hosted on Netlify, but it looks like it's using `adapter-static`. So it's really only dev mode that we're doing SSR at the moment

MacFJA:
There are several packages that offer this feature (and two of them ([storez](https://sveltesociety.dev/components#component-storez) and <span title="Yes, it's mine, but that's not the question here">[Persistent Store](https://sveltesociety.dev/components#component-Persistent%20Store) ℹ️</span>) are listed on this website, but [many more exists](https://www.npmjs.com/search?q=svelte%20localstorage)).

Why don't use one of them instead of reimplementing it?

ambarvm:
I tend to avoid packages for implementing small functionality to reduce bloat and I also wanted to try implementing a custom store myself.
Your persistent store package seems very nice and treeshakable. So I dont mind using that either 😄

kevmodrome:
> Oh, I guess it is a static export in production actually. I assumed we'd be using `adapter-netlify` since it's hosted on Netlify, but it looks like it's using `adapter-static`. So it's really only dev mode that we're doing SSR at the moment

Yeah only the PR previews and such are on Netlify, the actual site is on Cloudflare Pages. Do we want to discuss the idea of moving it to a Worker instead of pages? I don't know if there's anything that needs that at the moment though

PR #119, opened 2021-09-13
fixed bug on spacing

brittneypostma:
Trying this pr, had an error with package lock. This should just be a small style change to add media queries.


brittneypostma:
Okay, I'm kind of stuck here, the package-lock shouldn't be

PR #120, opened 2021-09-13
fix commit

brittneypostma:
Hoping this fixes the issue with the package lock.

### DIFF
--- a/src/routes/__layout.svelte
+++ b/src/routes/__layout.svelte
@@ -30,4 +30,10 @@
 	main {
 		padding: var(--s-10) var(--s-5) var(--s-20);
 	}
+
+  @media (min-width: 1280px) {
+    main {
+      padding: var(--s-20) var(--s-5);
+    }
+  }
 </style>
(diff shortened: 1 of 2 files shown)


brittneypostma:
Okay, @kevmodrome and @benmccann at a loss here. Accidentally commited the package lock and pushed this branch to fix that, but checks are still not passing. Any idea why? This should just be a minor change to add a couple media queries for the spacing on the home page.

src/routes/__layout.svelte
```diff
@@ -30,4 +30,10 @@
 	main {
 		padding: var(--s-10) var(--s-5) var(--s-20);
 	}
+
+  @media (min-width: 1280px) {
```
benmccann: lint is failing. I think you need to run `npm run format` to convert these spaces to tabs

brittneypostma: Oh gotcha let me fix that. I'm on windows, does that affect the tabs/spaces?

PR #124, opened 2021-09-17
Use location.search instead of $page.query

### LINKED ISSUE #123: Remove use of `$page.query`
@ABarnob looks like I was wrong in https://github.com/svelte-society/sveltesociety.dev/pull/108 that it was a SvelteKit bug and it was really a bug in the implementation, so maybe I shouldn't have merged it yet (more details available in the issues linked to from https://github.com/sveltejs/kit/issues/2444#issuecomment-921092592). Do you think you could fix it?

We'd either need to not prerender the page (e.g. switch from `adapter-static` to `adapter-netlify`) or update the dropdown in `onMount` using `location.search`). The latter seems potentially safer because it's a code change only and doesn't involve the possibility of needing to make any hosting changes, but @kevmodrome could weigh in if he feels differently

aantusahaa:
Resolves #123

### DIFF
--- a/src/routes/help/submitting.svelte
+++ b/src/routes/help/submitting.svelte
@@ -3,13 +3,11 @@
 	import components from '../components/components.json';
 	import templates from '../templates/templates.json';
 	import tools from '../tools/tools.json';
-	import { tick } from 'svelte';
-	import { page } from '$app/stores';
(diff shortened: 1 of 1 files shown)


benmccann:
thanks!!

PR #135, opened 2021-10-03
fix: add padding top in category tree

### LINKED ISSUE #102: Padding in CategoryTree
image
image

In both cases some extra margin in the top can be good

Both use the component `CategoryTree` which has `margin: 0 0 0 1.75rem;`

I suspect a top margin of 1rem should add sufficient white space

eriandev:
fixes #102 - Padding in CategoryTree


brittneypostma:
I'm going to allow this as a temporary fix until the component can be reworked. This will close #102.

PR #146, opened 2021-10-09
Merge staging into main

brittneypostma:


benmccann (approved):

benmccann:
Hmm. I just realized I think I screwed this up. I hit "squash and merge" since it was the default button, but that's probably not right. I don't think we'd want to squash all the commits together and lose the commit history. We'd probably want to do something more like a rebase. I can't fix it because the `main` branch is protected. @kevmodrome do you want to take a stab at fixing it or unprotect the `main` branch so that I can?

benmccann:
Also, it doesn't look to have redeployed the site. I don't know if I need to wait longer or if the deploy configuration for the `main` branch is broken

PR #168, opened 2021-11-09
issue #163

### LINKED ISSUE #163: component builder produces poorly formatted output
If you copy the output of https://sveltesociety.dev/help/submitting?type=component it will fail lint

It should output tabs instead of spaces, put the tags on a single line, etc.

srmullen:
Changes the json snippet to use tabs instead of spaces, and updates the styling on the pre tag so it looks the same as before (\t was indented 8 spaces without the style change).

This doesn't handle formatting the tags array because the correct formating is driven by prettier and depends on the length of the array. I think the best way to handle this is to add a formatting job to the github actions that runs before the linter. I can add that action with this pull request as well if desired.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,21 @@
+name: JSON Format
+
+on:
+  pull_request_target:
+    branches: [format]
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        ref: ${{ github.head_ref }}
+    
+    - name: Prettify code
+      uses: creyD/prettier_action@v4.0
+      with:
+        prettier_options: --write **/*.json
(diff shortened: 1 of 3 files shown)


benmccann:
We probably could detect the string length and whether it needed to be split into multiple lines, but this is a good start either way :-)

srmullen:
Thanks! My thought here was that it's probably best not to be reimplementing Prettier functionality because it wouldn't be tied to the lint/prettier config.

If implemented as a github action it could be setup to run only when the json files change. It would also be resilient against any other changes that a person would make when editing the file.

benmccann:
Can a Github action automatically update someone's fork of this repo though?

srmullen:
Hmm... Good question. Possibly using a `pull_request_target` event (https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target), but I don't know for sure.

PR #169, opened 2021-11-10
submit my project.

dufu1991:
submit my project by Svelte.


MacFJA:
Can you fix the indentation of the JSON ?
To be consistent with the rest of the file.

----

Also, I'm not sure if your project is in the right place. The **Templates** page is for packages/repositories/etc. that can be used to build a new new application based on it, and they are designed to ease the creation of new project.

You project seem more like a final product, not a starting point to create something new.
But I can be wrong, my chinese is non existent, and Google Translate is maybe misleading me in the understanding of your repo

PR #170, opened 2021-11-10
Feat/seo comp

johnwalshuk:
- Add consistent title format for all pages: `Title - Svelte Society`
- Fix issue where client page change did not update the page title
- Add human-friendly title to sub-pages of sections like: events and recipes
- Fix issue where previously set metadata can leak onto other pages
- Fix broken image


MacFJA:
Other PR around this subject have been done.

Can you update your PR to match to the current version of `staging`.
I think what you done here can fix the issue we have here https://github.com/svelte-society/sveltesociety.dev/pull/205#issuecomment-1022490843 and in #199

benmccann:
I overhauled the way that meta tags are calculated in order to fix that issue. I think it should hopefully work decently well now or at least a lot better. You'll have to rebase against the latest commit on `staging` now though

johnwalshuk:
Hey, im not exactly sure what you want me to do here, are the issues fixed in staging, or did you want me to alter the code so that I pass down `$page.stuff.metatags` to `<Seo />`?

benmccann:
The way I setup the metatags would allow you to have defaults in a layout and override in a sub-layout or page, which you couldn't do with the approach in this PR. But it turns out my approach is not compatible with MDSvex, so I think we should drop the code I added and use your code instead. Sorry for the difficulty. I'll take a look at updating your PR

benmccann:
Thanks for this!! A big improvement :smile:

PR #172, opened 2021-11-12
Update select component to unify styling, fix #100

### LINKED ISSUE #100: Select and ArrowButton Components
In the routes Templates, Components, and Tooling a new SvelteSelect package was added to replace the ArrowButtton component used before. The Components route still has one of the old ArrowButton in use. I have updated the styling as best I can to match, but even within the SvelteSelect package the styling is inconsistent, seen below in the Sorting dropdown. 

image

I suggest taking this package out and using vanilla html select element with the svelte stores and doing some custom styling that fits our design how this w3c post shows https://www.w3schools.com/howto/howto_custom_select.asp

mskocik:
I replaced the current select component by another, which provide somewhat easier way to styling and also works with mouse.
With this update also logic in given pages could be simplified in multiple places due the value select returns.

Update of svelte was also required.

Disclaimer: I am author of replacement select component

image


MacFJA:
The page `src/routes/help/submitting.svelte` should also be updated to use this component (to unify the select across the website).

And then the `svelte-select` package should be removed from the dependencies list in the `package.json`

mskocik:
Did that, anything else I should do?

MacFJA:
I see that the select for the package manager is still on the old style.

Also, is it possible to have the default value selected (like for the sorting order).

mskocik:
> I see that the select for the package manager is still on the old style.

I am not sure what you mean by old style. You mean input size?

PR #173, opened 2021-11-13
Fix typo on cheatsheet

OscarAblinger:
Fixing a minor typo that I found on the cheatsheet.

I also removed the `()` around the `console.log`. AFAIK they shouldn't make a difference.


OscarAblinger:
Oh and quick note: the linked repl does not share the same typo.

PR #175, opened 2021-11-14
Replaced deprecated package @cloudedots/svelte-forms with svelte-form-validation

DhyeyMoliya:
Replaced a deprecated package [@cloudedots/svelte-forms](https://www.npmjs.com/package/@cloudedots/svelte-forms) in favour of a new package [svelte-form-validation](https://www.npmjs.com/package/svelte-form-validation).

### DIFF
--- a/src/routes/components/components.json
+++ b/src/routes/components/components.json
@@ -1,5 +1,15 @@
 [
 	{
+		"title": "svelte-form-validation",
+		"url": "https://github.com/DhyeyMoliya/svelte-form-validation",
+		"npm": "svelte-form-validation",
+		"description": "Svelte Form Validation Library",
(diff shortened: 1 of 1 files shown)


MacFJA:
I'm not in favor of removing a component. It still exist so it can be still be used.
I would rather prefer to update the name and/or description to reflect that the component is deprecated (and it has a suggested replacement).

benmccann (approved):
personally I think it's fine to remove a deprecated package if the package author is the one suggesting it. DhyeyMoliya appears to be the author in this case. It seems like a better experience for our users not to have to navigate through a deprecated package to find the current one and the old one didn't have many users so I don't think anyone would miss it

PR #177, opened 2021-11-16
Add @budgetdraw/sveltekit-cloudflare-adapter

benmccann:


benmccann:
oh, that's annoying. this seems to be failing because someone checked in code that doesn't lint. see https://github.com/svelte-society/sveltesociety.dev/pull/178 for the fix

PR #193, opened 2021-12-13
Update Meteor bundler.

wizonesolutions:
The official Meteor tutorial has switched to using the Meteor package `zodern:melte`, so I'm proposing an update here.

### DIFF
--- a/src/routes/tools/tools.json
+++ b/src/routes/tools/tools.json
@@ -73,15 +73,6 @@
 	},
 	{
 		"addedOn": "2021-08-09T10:14:05.723Z",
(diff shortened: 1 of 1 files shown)


Ennoriel:
@wizonesolutions the build is failing. There is a trailing comma at the end of json file (which would be right in a js object 😀)

wizonesolutions:
Ack, that always seems to happen! I've moved to using SvelteKit for now, as the Meteor bundler is too unmaintained. I figure someone can make their own PR if they want this later, but won't be following up myself.

PR #217, opened 2022-01-22
Github Action to update stars

### LINKED ISSUE #16: Periodically update components and boilerplates data
We should run a script that periodically that updates the site's data. E.g. number of stars that a repo has, last date the repo was updated, etc. Also, I noticed an archived repo on the components page in https://github.com/svelte-society/sveltesociety-2021/pull/15, which is something we could detect as well

The main Svelte site gets info from GitHub [to get the list of contributors](https://github.com/sveltejs/svelte/blob/master/site/scripts/get-contributors.js).

**GraphQL API**

It looks to me like this has a much higher rate limit than the REST API. However, authentication is required.

You can get 500,000 nodes in one call. It has a rate limit of 5,000 points per hour. I think that translates into being able to load 500,000 repos / hr since the things we care about like `pushedAt` and `stargazerCount` are on the same node. https://docs.github.com/en/graphql/overview/resource-limitations

**REST API**

Info about the Github API rate limit is here: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting
* Unauthenticated - 60 requests / hr
* GitHub Actions - 1,000 requests / hr
* Logged in users - 5,000 requests / hr

It looks like you can only get one repo at a time: https://docs.github.com/en/rest/reference/repos#get-a-repository

### LINKED ISSUE #368: Github star count should be automatically sync with GitHub
Currently the star count of a component is hard coded. it should be pull from GitHub.

MacFJA:
Github Workflow and Action to update star cont for Github and Gitlab repositories.

Limitation: If the repository change name, the star count will not be found.

Related to #16 #368


.github/actions/update-stars/main.cjs
```diff
@@ -0,0 +1,108 @@
+const core = require('@actions/core');
```
benmccann: could you make this a `.js` file instead of `.cjs` file?

.github/actions/update-stars/main.cjs
```diff
@@ -0,0 +1,108 @@
...
+const axios = require("axios");
+
+const ghGraphQlUrl = 'https://api.github.com/graphql'
+const gitlabGraphQlUrl = 'https://gitlab.com/api/graphql'
+const githubNameRegexp = new RegExp('https://github.com/([a-zA-Z0-9][a-zA-Z0-9-]{0,38}/[a-zA-Z0-9._-]{1,100})')
+const gitlabNameRegExp = new RegExp('https://gitlab.com/([\\w-]+/[\\w-]+)')
+
+async function doGraphQlQuery(url, query, headers = {}) {
+    try {
+        let fetchResponse = await axios({
```
benmccann: shouldn't need to use `axios`. can just use the `fetch` built into node if you're on node 18 (https://nodejs.org/api/globals.html#fetch) - or possibly `undici`, which is the `fetch` that is bundled with node, if we need to support older versions, but that shouldn't be necessary

MacFJA: Sadly, GitHub action only allows Node 12 and Node 16.
(And the doc don't even mention Node 12).

I switch to `node-fetch`, which is more close to the native `fetch` (less work when Node 18 will be available)

benmccann: I think you just need to add https://github.com/actions/setup-node. E.g. in https://github.com/svelte-society/sveltesociety.dev/pull/424 you used Node 18

MacFJA: A workflow can use `setup-node` (or whatever they need, as they run on a full container), but an action is a bit more limited:

- https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action
- https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsusing-for-javascript-actions

---

Maybe there are other method to make an action, or I can transform the action into a workflow step (If everything is possible as a workflow step) :thinking:

benmccann: Ah, okay. In that case I think we should use `undici` rather than `node-fetch`. It's exactly the same as what's included in Node so later when we use a newer Node we can just drop the dependency and everything will keep working exactly the same

.github/actions/update-stars/main.js
```diff
@@ -0,0 +1,131 @@
+import core from '@actions/core';
+import { readFileSync, writeFileSync } from 'fs';
```
```suggestion
import { readFileSync, writeFileSync } from 'node:fs';
```

.github/workflows/update-stars.yml
```diff
@@ -0,0 +1,35 @@
+name: Update stars count
+
+on:
+  schedule:
+    - cron: '0 0 * * 0' # Every Sunday at 00:00
```
benmccann: Do you think we'd get tired of having to merge a PR every week? I doubt things change that much. Maybe once a month would be enough?

.github/workflows/update-stars.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup node with version 16
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
```
benmccann, line 16:
```suggestion
      - name: Setup Node.js
        uses: actions/setup-node@v3
          node-version: 20
          cache: 'npm'
```

.github/actions/update-stars/main.js
```diff
@@ -0,0 +1,131 @@
+import core from '@actions/core';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { fetch } from 'undici';
```
benmccann, line 3: actually, let's just set the node version to 20 in the CI. then `fetch` will be available as a global. https://nodejs.org/dist/latest-v20.x/docs/api/globals.html#fetch
```suggestion
```

.github/actions/update-stars/action.yml
```diff
@@ -0,0 +1,9 @@
+name: Update stars
+description: Update Github stars count
+inputs:
+  token:
+    description: GitHub token for GraphQL calls
+    required: false
+runs:
+  using: node16
```
benmccann, line 8:
```suggestion
  using: node20
```

package.json
```diff
@@ -39,6 +40,7 @@
 		"svelte-select": "^4.4.0",
 		"tslib": "^2.6.0",
 		"typescript": "^5.1.6",
+		"undici": "^5.22.1",
```
benmccann, line 43:
```suggestion
```

benmccann:
forget my comments. I forgot our earlier discussion :laughing:

benmccann:
@MacFJA I just tried to run this locally to do a one-off update before the first scheduled job runs and it seems like it didn't work or maybe I didn't do it correctly. 
```
$ node .github/actions/update-stars/main.js 
Start
	Github: 0 repositories (0 stars)
	Gitlab: 3 repositories (21 stars)
Done
```

Should that work and is it a bug that it found 0 repos for GitHub?

benmccann:
@MacFJA stars have disappeared from a lot of projects on https://sveltesociety.dev/components. I think it happened when I merged this PR. Maybe you could take a look?

PR #234, opened 2022-02-18
Addition of an application to manage expenses

ln-dev7:


MacFJA:
I'm not sure that what you want to add is a _template_.

To me it look more like a final product.

----

- Do you expect people to copy your repository content to create something else ?
- Does your repository offer a solution to a common problem ?
- Does your repository is an integration of some tools / packages / external solution ?

benmccann:
I agree this is not a template. A better place to showcase this might be Svelte's monthly newsletter

PR #241, opened 2022-02-28
Add lib svelte-lazy-loader to components

sawyerclick:
Add svelte-lazy-loader to the components page. It's a library built with SvelteKit that, if native lazy-loading isn't supported by the browser, uses a shared IntersectionObserver instance to load images. The API is identical to HTLMImageElement. More components are forthcoming in the library as well.

### DIFF
--- a/src/routes/components/components.json
+++ b/src/routes/components/components.json
@@ -1,5 +1,17 @@
 [
 	{
+	    "title": "svelte-lazy-loader",
+	    "url": "https://github.com/sawyerclick/svelte-lazy-loader",
+	    "description": "A lazy-loader component library for Svelte",
+	    "npm": "svelte-lazy-loader",
+	    "addedOn": "2022-02-27",
+	    "tags": [
+		"images",
+		"components and libraries"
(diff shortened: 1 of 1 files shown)


src/routes/components/components.json
```diff
@@ -1,4 +1,16 @@
...
+	    "url": "https://github.com/sawyerclick/svelte-lazy-loader",
+	    "description": "A lazy-loader component library for Svelte",
+	    "npm": "svelte-lazy-loader",
+	    "addedOn": "2022-02-27",
+	    "tags": [
+		"images",
+		"components and libraries"
+	    ],
+	    "stars": 0
+	}
```
MacFJA: There is a missing comma here
```suggestion
	},
```

MacFJA (changes requested):

MacFJA:
I see mix usage of spaces and tabs.
The file only allow tabs.

You can run the command `npm run format` to automatically fix them

MacFJA (approved):

PR #247, opened 2022-03-10
Prevent  `undefined` from appearing in filter dropdowns

### LINKED ISSUE #233: `undefined` component category
Screenshot from 2022-02-15 15-55-57

nvlgzr:
This should fix #233, while also providing a bit more test coverage. :)

The most salient feature is that `extractUnique` transforms missing fields in the data into empty strings instead of `undefined`.


src/lib/utils/extractUnique.ts
```diff
@@ -2,8 +2,14 @@ export const extractUnique = (
 	source: Array<Record<string, unknown>>,
 	field: string
 ): Array<Record<'label' | 'value', unknown>> => {
-	const extracted = Array.from(new Set(source.map((item) => item[field]).flat()));
```
vhscom, line 5: Haven't tried it but seems like a case for nullish coalescing, e.g.

```ts
const extracted = Array.from(new Set(source.map((item) => item[field] ?? '').flat()));
```

nvlgzr: Oh, sure. Done!

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+	"editor.insertSpaces": false
+}
```
MacFJA: I'm not sure this configuration file should be here. :thinking: 

It's configuration about your personal choice, and not project choice.
Moreover it's a development program specific configuration (VSCode)

vhscom: VSCode slips these in almost at easily as macOS does `.DS_Store` files. Might be worth adding to `.gitignore` when this file is removed.

nvlgzr: Sorry for the delayed response, @MacFJA. I've reverted the commit. :)

That said, I disagree with your middle sentence:

> It's configuration about your personal choice, and not project choice.

The choice to use tabs vs spaces should very much be a project choice since otherwise you run the risk of every change affecting every line because the whitespace keeps getting replaced every time people with opposite _personal_ preferences edit the same file.

I can see the future that you want to avoid, where contributors end up spamming the project root with various editor configs. Might something like https://editorconfig.org/ be worth adopting?

mapokapo:
I just visited the site and after selecting "official" in the *tags* filter the entire filters section freezer and stops working. It produces the following error message:
```
index.svelte-45688c86.js:1
Uncaught (in promise)
TypeError: Cannot read properties of undefined (reading 'includes')
    at index.svelte-45688c86.js:1:6070
    at Array.filter (<anonymous>)
    at P (index.svelte-45688c86.js:1:6058)
    at index.svelte-45688c86.js:1:6719
    at Array.filter (<anonymous>)
    at Ae.i.$$.update (index.svelte-45688c86.js:1:6554)
    at Sl (vendor-d78fc752.js:4:1688)
    at Hn (vendor-d78fc752.js:4:1438)
```

Please merge this as fast as possible :D

benmccann:
I merged this and redeployed the site. Component tags still aren't working for me, which is odd because they looked fine on staging

PR #248, opened 2022-03-12
update components.json : added new snippet

ghost:


MacFJA:
There is a mix usage of spaces and tabs in the files.

Only tabs are allowed.

You can run the command `npm run format` in the project to normalize the file

PR #249, opened 2022-03-12
update components.json : added new snippet

ghost:


MacFJA (commented):
The `package-lock.json` shouldn't be a part of your PR.

The changes inside it is not directly link to what you want: adding a new component.

In a way, it's won't do anything bad, but, **IMO**, the PR will look a bit more cleaner without _extra_ files

PR #252, opened 2022-03-16
add maverick-dev-55/svelte-savable

mavdotj:


mavdotj:
I don't endorse using this package. try `@macfja/svelte-persistent-store`

PR #255, opened 2022-03-23
Add `svelte-stopwatch`

erokar:


src/routes/components/components.json
```diff
@@ -1,4 +1,20 @@
...
+	    "description": "A Svelte timer component with a countdown option and events for start, stop and reset of the timer.",
+	    "npm": "svelte-stopwatch",
+	    "addedOn": "2022-03-23",
+	    "category": "User Interaction",
+	    "tags": [
+		"time and date",
+		"components and libraries",
+		"notifications",
+		"timer",
+		"countdown"
```
benmccann: i'm not sure we should be adding so many new tags

erokar: Have reduced the number of tags, feel free to edit the list as you see fit.

PR #259, opened 2022-04-12
add books section

tanhauhau:
add a section to list books about svelte

### DIFF
--- a/src/lib/components/layout/Header.svelte
+++ b/src/lib/components/layout/Header.svelte
@@ -10,7 +10,8 @@
 	const linksRight = [
 		['/recipes', 'recipes'],
 		['/cheatsheet', 'cheat sheet'],
-		['/events', 'events']
+		['/events', 'events'],
+		['/resources', 'resources']
 	];
 </script>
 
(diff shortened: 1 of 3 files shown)


pngwn:
Perhaps this should be called `Resources` and could be more general purpose?

benmccann:
we should probably also move [the video faq](https://svelte.dev/faq#are-there-any-video-courses) here as well

I think this is too many menu items unfortunately especially since it ends up being off center. I don't have any great suggestions for how to handle it. We probably need an entirely new menu implementation. In the meantime, maybe we can remove the Cheat Sheet menu item and link to the Cheat Sheet from Recipes?

benmccann:
@MacFJA @brittneypostma @kevmodrome do y'all have thoughts about how to handle this PR which adds another menu item? I listed one idea in my last comment, but wanted to get another opinion if possible before moving ahead with any decision

brittneypostma:
Isn't @kevmodrome working on entirely new site design as well?

brittneypostma:
In my opinion, if anything is going on this older site, resources should stay and have cheat sheet just go under resources.

PR #264, opened 2022-04-14
Updated template.json

timoyo93:
Added a new template to the template.json I created a while ago and find it very useful


MacFJA:
Mix usage of tabs and spaces.

JSON file are indented only by tabs.

You can run `npm run format` to fix it automatically.

PR #274, opened 2022-05-19
added svelte-brick-gallery

anotherempty:


src/routes/components/components.json
```diff
@@ -2546,5 +2546,15 @@
...
 		"stars": 0
+	},
+	{
+		"title": "svelte-brick-gallery",
+		"url": "https://github.com/anotherempty/svelte-brick-gallery",
+		"description": "A masonry-like image gallery component for svelte",
+		"npm": "svelte-brick-gallery",
+		"addedOn": "2022-05-19",
+		"category": "User Interaction",
+		"tags": ["image", "gallery", "masonry"],
```
MacFJA: The tags `gallery` and `masonry` are new.

I'm not sure about the need to create new ones.
Any thought about it @kevmodrome @benmccann ?

benmccann: I would avoid it. They don't seem like they'll be heavily used

PR #284, opened 2022-07-25
Add svelte-vite-typescript-tailwindcss-template

skshahriarahmedraka:
added a template , svelte-vite-typescript-tailwindcss-template


benmccann:
I'm not really sure the point of a template for this when we have adders. It seems destined to get outdated

PR #293, opened 2022-08-14
Use external lib for search

### LINKED ISSUE #100: Select and ArrowButton Components

### LINKED ISSUE #292: Empty and unnecessary "Unclassified" section
https://sveltesociety.dev/tools/

MacFJA:
Replace #181 
Fix #100 
Fix #292

---

- Better search (use tool design for it instead of one made by hand)
- Simplify Svelte component readability (IMO)
- Fix error with unclassified/uncategorized item

---

- [x] Update Components
- [x] Update Templates
- [x] Update Tools


src/lib/components/Search.svelte
```diff
@@ -0,0 +1,101 @@
+<script lang="ts">
+	import type { Configuration } from 'itemsjs';
+	import itemsjs from 'itemsjs';
+	import { createEventDispatcher } from 'svelte';
+	import Select from '$lib/components/Select.svelte';
+
+	const dispatch = createEventDispatcher();
+
+	export type Facet = {
```
benmccann: does this need to be exported?

benmccann (commented):
Thanks for fixing this!

it's a bit hard to differentiate between "uncategorized" and empty input in the category drop down

I think we could simplify the sort. E.g. we could just have "stars" instead of "stars asc" and "stars desc" as there's no real utility in having "stars desc". I'm not sure we need both the "asc" and "desc" for any of them

I wonder if we need the weight of a new library? Any idea what was wrong before and if it'd be hard to just fix it?

src/lib/components/Search.svelte
MacFJA: You're right, no need. Pages don't use Typescript (ne `<script lang="ts">`)

MacFJA:
> it's a bit hard to differentiate between "uncategorized" and empty input in the category drop down

Yes, and I also see that **Tools** miss the `Uncategorized`

> I think we could simplify the sort. E.g. we could just have "stars" instead of "stars asc" and "stars desc" as there's no real utility in having "stars desc". I'm not sure we need both the "asc" and "desc" for any of them

Good point

> I wonder if we need the weight of a new library? Any idea what was wrong before and if it'd be hard to just fix it?

What do you mean ?

benmccann:
I'd slim the sorting options down some more. Like: "Stars", "Name", "Most recent", "Oldest". I don't really know what use case would have people looking for components starting with Z. Don't need to say "first" as it's implied

When I said I wonder if we need the weight of a new library, I meant do we need the extra KB it will add. I'm not overly concerned about this as it's more important just to get it to work, but without looking at the implementation much myself I was a little surprised it'd warrant an external library.

Thanks again for fixing this!

MacFJA:
> I'd slim the sorting options down some more. Like: "Stars", "Name", "Most recent", "Oldest". I don't really know what use case would have people looking for components starting with Z. Don't need to say "first" as it's implied

Done


> When I said I wonder if we need the weight of a new library, I meant do we need the extra KB it will add. I'm not overly concerned about this as it's more important just to get it to work, but without looking at the implementation much myself I was a little surprised it'd warrant an external library.

The old (current) implementation of the search feature, is not very optimized.
While I appreciate the time I spend and the challenge I tackle to make it, I'm obliged to recognize that the implementation is far from ideal.
The fulltext search is pretty naive, the facets is quite heavy and hard to read, the sorting is limited.

There are better solution that the one I wrote. :sweat_smile:

src/routes/_searchableJson.svelte
```diff
@@ -0,0 +1,97 @@
...
+
+<SearchLayout title={displayTitle}>
+	<section slot="controls" class="controls">
+		<div class="inputs">
+			<Search
+				{data}
+				dataDefault={{ category: '' }}
+				bind:query={searchValue}
+				sortableFields={[
+					{ identifier: 'addedOn', title: 'Last added', ascending: false },
```
benmccann:
```suggestion
					{ identifier: 'addedOn', title: 'Most recent', ascending: false },
```

src/routes/_searchableJson.svelte
```diff
@@ -0,0 +1,97 @@
...
+					{ identifier: 'addedOn', title: 'Oldest', ascending: true },
+					{ identifier: 'title', title: 'Name', ascending: true },
+					{ identifier: 'stars', title: 'Most stars', ascending: false }
```
benmccann:
```suggestion
					{ identifier: 'stars', title: 'Stars', ascending: false }
```

benmccann (commented):
It doesn't necessarily have to be part of this PR, but I noticed that the text for the "Tags" field is quite high rather than being vertically centered

Screenshot from 2022-09-01 12-36-59

MacFJA:
> It doesn't necessarily have to be part of this PR, but I noticed that the text for the "Tags" field is quite high rather than being vertically centered
> 
> Screenshot from 2022-09-01 12-36-59

Indeed not the purpose of this PR.

I also have this issue, but only for a few seconds on the initial load.
(But on the production, I have it all the time)

benmccann:
Looks like this PR will need a rebase now that your migration PR has been merged

MacFJA:
Rebase done :)

PR #297, opened 2022-08-22
Remove unused text in unit testing page

marktnoonan:
Unless something is truly "to be written" here. Excited to be learning more about testing Svelte components!


kevmodrome:
Closing this since we're moving to a non git based system for content.

PR #300, opened 2022-08-29
Fix node version of the CI

MacFJA:
Since the update of the version of SvelteKit, the Lint CI action is failing due to the fact that SvelteKit (or one of its dependencies) require Node 16.x to work


benmccann (approved):
lgtm. although now the lint job is failing due to formatting in one of the files. not sure if you want to fix that as part of this PR or separately

PR #309, opened 2022-09-13
update `typesafe-i18n` entry

ivanhofer:


benmccann:
Removed from tools in favor of components entry

PR #314, opened 2022-09-15
update link for CZ/SK society.

Mlocik97:


benmccann:
both links appear to work to me. does it matter which we use?

Mlocik97:
old one shows me "15 usages".

benmccann:
What do you mean by that? I see the same thing with either link. Here's a screenshot

Screenshot from 2022-09-20 10-36-30

Mlocik97:
I mean, if 15 people accept that invite, then it will be invalid? 16th one will be unable to connect via that link?

PR #321, opened 2022-10-04
Adding an animation on hover to list titles

fedeperin:
Added an animation on hover to the list title (that is a link) at Recipes page
Recording_1664925176091-_1_-_1_

### DIFF
--- a/src/routes/recipes/+page.svelte
+++ b/src/routes/recipes/+page.svelte
@@ -152,6 +152,21 @@
 		top: 4px;
 	}
 
+	.list-title::before {
+		content: '';
+		background: var(--black);
+		opacity: 0.8;
+		height: 0px;
+		width: 100%;
+		position: absolute;
+		bottom: 5px;
+		transition: all 0.05s ease;
+	}
+
+	.list-title:hover::before {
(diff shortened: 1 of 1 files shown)


src/routes/recipes/+page.svelte
```diff
@@ -152,6 +152,21 @@
...
 
+	.list-title::before {
+		content: '';
+		background: var(--black);
+		opacity: 0.8;
+		height: 0px;
+		width: 100%;
+		position: absolute;
+		bottom: 5px;
+		transition: all 0.05s ease;
```
MacFJA:
```suggestion
		transition: height 0.05s ease;
```
The property that you want to animate is `height` not all possible properties.
To avoid side effect, it's better to target only the thing that you are changing.

src/routes/recipes/+page.svelte
```diff
@@ -152,6 +152,21 @@
 		top: 4px;
 	}
 
+	.list-title::before {
```
MacFJA, line 155: It's a matter of preferences, but I would use `after` instead of `before`, because the line you add is **after** the content

PR #324, opened 2022-10-09
Add SvelteKit Authentication Example

nstuyvesant:
Add SvelteKit Authentication Example

### DIFF
--- a/src/routes/components/components.json
+++ b/src/routes/components/components.json
@@ -1,5 +1,15 @@
 [
 	{
+		"title": "SvelteKit Authentication Example Project",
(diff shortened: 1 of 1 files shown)


MacFJA:
I wonder if it's a component or a template :thinking:

nstuyvesant:
Really a template

MacFJA:
> Really a template

If it's a template, then you should add it in the [`templates.json`](/svelte-society/sveltesociety.dev/blob/staging/src/routes/templates/templates.json) file, not in `components.json`

PR #331, opened 2022-11-02
Add `sveltekit-adapter-firebase`

simonnepomuk:


benmccann:
What's the difference between this package and the existing https://github.com/jthegedus/svelte-adapter-firebase package?

simonnepomuk:
> What's the difference between this package and the existing https://github.com/jthegedus/svelte-adapter-firebase package?

The adapter you linked is not compatible with the current version of SvelteKit. Also it requires more configuration effort and is harder to set up.

PR #364, opened 2023-01-24
add simple-ui-components-in-svelte

tborychowski:


MacFJA:
There were an issue with the preview deployment, which have been solved.
Can you rebase your branch ?

---

Also the `lint` task failed.
You can run `npm run format` to fix it

tborychowski:
Updated my fork with the latest, thanks.
The `npm run format` doesn't change anything.

src/routes/components/components.json
```diff
@@ -2827,5 +2827,15 @@
...
 		"stars": 0
+	},
+	{
+		"title": "simple-ui-components-in-svelte",
+		"url": "https://perfect-things.github.io/simple-ui-components-in-svelte/",
+		"description": "A set of simple & flexible UI components",
+		"npm": "simple-ui-components-in-svelte",
+		"addedOn": "2023-01-24",
+		"category": "Forms & User Input",
+		"tags": ["forms", "UI"],
```
MacFJA: The tag UI doesn't exist

tborychowski: That's a pity. It's a very useful tag 😉 
Removed.

MacFJA: New tag can be added, but they need to be thoroughly thinked.
If a new tag is added, all other components need to be review to check if they can have it too.
And if almost all of them have it, the tag don't have a lot of meaning for the end user that want to filter (i.e. if adding the filter remove 2 items in a list of 300, it's worth nothing). And in the contrary, if a tag match almost nothing, it's either a very good filter or a terrible one.

PR #365, opened 2023-01-28
only warn on missing ids

gtm-nayan:
Mandatory: this is not a permanent solution.

The recipes pages have anchor links but don't actually render the ids, and the components page has way too many links with encoded ids which needs to handled in Kit's prerenderer. For now, this should make the builds succeed.


benmccann:
thank you for finding this!! I think I've gotten all the links fixed, so we won't need to ignore them

PR #374, opened 2023-02-12
Update stars

nstuyvesant:


MacFJA:
It's the 8th PR you open to update the star count.
I know that the fact that stars not being updated I far from be ideal, but it should have a way to update star more globally.

There are several issues/PR about it:
- #368
- #16
- #217

Please add a comment on any of them to raise the awareness about the need of a "definitive" solution to update stars count.
You can also [open a new discussion](https://github.com/svelte-society/sveltesociety.dev/discussions).

---

_P.S. I'm not against merging your PR, but I prefer a fair approach to stars of **all** items, and to have a tool/process that will do it automatically (which can avoid errors, conflict, unnecessary commit in the git log)_
_And to be totally honest, I will probably merge it in a few days (1 or 2)... unless someone else do it before me :sweat_smile:_

nstuyvesant:
I understand what you're saying about the stars. It would be good to get these in an automated way. I would like to see my description update make it to the production site though. What's on staging is good.

MacFJA:
The production/live website update have been requested: 
- #375

PR #380, opened 2023-02-24
Add sveltepress

Blackman99:


src/routes/templates/templates.json
```diff
@@ -897,5 +897,15 @@
 		"category": "Svelte",
 		"tags": ["storybook", "typescript"],
 		"stars": 0
+	},
+	{
+		"title": "Sveltepress",
+		"url": "https://github.com/SveltePress/sveltepress",
+		"description": "A markdown centered site build tool with full power of sveltekit.",
+		"npm": "@sveltepress",
```
MacFJA: This is not a valid NPM package name

MacFJA (changes requested):

Blackman99:
The valid npm command is `npm create @sveltepress` not `npm install @sveltepress`
However there's no such an option

src/routes/templates/templates.json
Blackman99: The valid npm command is `npm create @sveltepress` not `npm install @sveltepress`
However there's no such an option

MacFJA: Maybe we should change the NPM/Yarn/PNPM command to `create` on the templates pages ? :thinking: 

I have to check if this change doesn't create a side effect for the existing templates

Blackman99: That would be great!

lachlancollins (approved):

PR #389, opened 2023-03-14
Update components.json

N00nDay:
Updated stars for STWUI


MacFJA:
Due to an error cause by fffedf4 the CI was broken.
Can you please rebase your branch with `staging` to get the fix.

benmccann:
closing due to inactivitiy. plus there's a PR pending to add a script to update all stars, which would be better than handling these as one offs

PR #393, opened 2023-03-19
Alternative solution for `components.json`, `templates.json` and `tools.json`.

ericfennis:
## Why

I noticed there are a lot of PRs having merge conflicts because another PR is merged with changes in `components.json`, `templates.json` or `tools.json`. Resulting in asking the contributors every time to rebase their PR. This is not very efficient and maintaining this can be a pain.

## Solution
Move each "repo" to its own JSON files.

To test this I started by moving the last added one to a JSON file. I think we could easily create a script for migrating all items in the JSON files to their own JSON files. But to make sure it is working with the current open PRs I kept the current JSON files for now.

### Notes
The current `updateStars.js` is not working with this setup. 
I recommend removing this script and extending the "page load data" scripts I created for tools, templates, and components. We could fetch the GitHub stars there. So they will be updated every build. This will fixes #217


MacFJA:
I added your idea to the discussion #318

kevmodrome:
Closing this as we will be using a completely different system in the v2 version of the site.

PR #394, opened 2023-03-20
Sveltyped

MrsHorrid:
sveltyped is a component library that allows you to easily add a typewriter effect


stephane-vanraes:
tags: add `components and libraries`, remove `templates` (it's not a template)
also fix the formatting please

MrsHorrid (commented):
changed what was asked of me

PR #398, opened 2023-03-30
add svelte-pilot

jiangfengming:
https://github.com/jiangfengming/svelte-pilot


lachlancollins (commented):
Already added in #498

PR #401, opened 2023-04-12
Adding Apity to components

lietu:
Apity is a typed API client for Svelte and SvelteKit working with TypeScript types generated from an OpenAPI spec.


lietu:
"Needs rebase"? Um .. Yeah, you waited 2 months, so of course it does. Go ahead and rebase.
image

PR #427, opened 2023-07-17
feat: Stricter types and json file validation

lachlancollins:
This PR makes a bunch of small tweaks that aim to bring this project closer to type safety:
- Added a zod schema to parse json files during build step (this is currently quite lenient with optional fields)
- Added build to the PR GitHub action (fails if json parsing is invalid)
- Typed `SearchableJson.svelte` based on the schema type
- Replaced `categoryId`s with slugs generated by `@sindresorhus/slugify`
- Fixed some small type errors (e.g. `$page.url.host`)
- Moved some shared libs out of routes folder


.github/workflows/pr.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          node-version: '18.x'
+          registry-url: https://registry.npmjs.org/
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install
+      - name: Run lint
+        run: pnpm run lint
+      - name: Run build
+        run: pnpm run build
```
MacFJA: Isn't `pnpm run check` enough for this ?

_(Should be faster than a build, but a build is not that long either)_

MacFJA (approved):

.github/workflows/pr.yml
lachlancollins: I wish! Only build will run the Zod schemas (runtime), and also svelte-check will still fail as not all type errors are fixed.

lachlancollins: The alternative I experimented with was renaming it from a json file to something like `data.js` then typing the file directly. SolidJS does a similar thing: https://github.com/solidjs/solid-site/blob/main/src/pages/Resources/Utilities.data.ts

MacFJA: What about a small script that run: 
```ts
import { entrySchema } from './src/lib/schema';
import components from './src/routes/components/components.json';
import tools from './src/routes/toold/tools.json';
import templates from './src/routes/templates/templates.json';

Object.entries({components, tools, templates}).forEach(([name, data]) => {
    console.log(`Validating ${name}...`)
    entrySchema.parse(data);
});
```
It will fail on the first file that is incorrect

lachlancollins: Ooo I like that!

package.json
```diff
@@ -9,13 +9,14 @@
 		"build": "vite build",
 		"preview": "vite preview",
 		"check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
-		"test": "vitest",
+		"test": "vitest run && node scripts/validateData.js",
```
MacFJA: Do we test (as Unit testing, functional testing) the data, or we lint (as check if it's matching our quality criteria) the data ?

For me, it's the latter, but I'm open to discussion

src/lib/pageList.ts
```diff
@@ -1,13 +1,13 @@
-export async function getPages(
-	metaGlob: Array<string>
-): Promise<Array<{ [key: string]: string; filename: string; path: string; title?: string }>> {
+export const getPages = async (metaGlob: Record<string, () => Promise<unknown>>) => {
```
MacFJA: Why `unknown` ? Look like we know the final structure (that we use*):
```suggestion
export const getPages = async (metaGlob: Record<string, () => Promise<{metadata: { title: string; layout: string; date: string }}>>) => {
```
It will also remove the need to type the data at lines 4 to 6

*Indeed, it's a partial structure, but we only need to know the part we use

MacFJA (commented):
- Why the change from `src/routes/recipes/+layout.ts` to `src/routes/recipes/+layout.js` (TS to JS) ?
- Why the change from `src/routes/events/+page.ts` to `src/routes/events/+page.js` (TS to JS) ?
- Why the change from `src/lib/schema.ts` to `src/lib/schema.js` (TS to JS) ?

lachlancollins:
> * Why the change from `src/routes/recipes/+layout.ts` to `src/routes/recipes/+layout.js` (TS to JS) ?
> * Why the change from `src/routes/events/+page.ts` to `src/routes/events/+page.js` (TS to JS) ?
> * Why the change from `src/lib/schema.ts` to `src/lib/schema.js` (TS to JS) ?

For `+page`/`+layout`, it doesn't really matter, it was just personal preference / SvelteKit pushes JSDoc typing pretty strongly (e.g. default on docs).

For `schema.js`, it needs to be JS to avoid additional complexity since it is imported from `validateData.js` and runs in Node.

MacFJA:
> For `+page`/`+layout`, it doesn't really matter, it was just personal preference / SvelteKit pushes JSDoc typing pretty strongly (e.g. default on docs).

The @sveltejs team have their belief on usage of Typescript vs. JSDoc (they have great article about it).
But it's not because they do it that it's the only way of doing it.

That's said, on this project, we have both JS and TS mix up without any rules.
So for the time being, I'm okay with your modifications

MacFJA:
@benmccann Your thought about this PR ?
Looks good to me, but as it contains lots of _"structural"_ changes, I prefer to have a second approval on it.

.eslintrc.cjs
```diff
@@ -28,6 +28,8 @@ module.exports = {
 		}
 	],
 	rules: {
+		'@typescript-eslint/ban-ts-comment': 'off',
+		'@typescript-eslint/no-explicit-any': 'off',
```
benmccann: why did this have to be added?

.github/workflows/pr.yml
```diff
@@ -1,18 +1,26 @@
...
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  lint:
+  pr:
+    name: PR
```
benmccann: PR feels like too generic of a name and doesn't describe what the job is actually doing. going back to a name like `lint` would be much better in my opinion

src/routes/events/+page.js
```diff
@@ -1,8 +1,8 @@
 import { error } from '@sveltejs/kit';
-import type { PageLoad } from './$types';
-import { getPages } from '../pageList';
+import { getPages } from '$lib/pageList';
 
-export async function load(): Promise<PageLoad> {
+/** @type {import('./$types').PageLoad} */
```
benmccann: I think you can get all the type benefits without this line: https://svelte.dev/blog/zero-config-type-safety
```suggestion
```

src/routes/recipes/+layout.js
```diff
@@ -1,28 +1,29 @@
 import { error } from '@sveltejs/kit';
 import '$styles/highlight.css';
-import type { PageLoad } from './$types';
-import { getPages } from '../pageList';
+import { getPages } from '$lib/pageList';
 
-export async function load(): Promise<PageLoad> {
+/** @type {import('./$types').LayoutLoad} */
```
benmccann:
```suggestion
```

.eslintrc.cjs
lachlancollins: Oops sorry, I was going for zero type errors and passing `pnpm check` at one point - I've removed this now.

.github/workflows/pr.yml
lachlancollins: Changed back :)

src/routes/events/+page.js
lachlancollins: I forgot about this, thanks!

PR #436, opened 2023-08-13
Update script to paginate request + init stars

MacFJA:
- Script:
  - Update the script to paginate request to avoid timeout from GitHub API.
  - Sort result to ease stars file review and merge
- Provide a base stars count for all repos


.github/actions/update-stars/main.js
```diff
@@ -41,6 +41,11 @@ function gatherUrls() {
 
 // Github
 
+/**
+ * Get all GitHub repositories
+ *
```
```suggestion
```

.github/actions/update-stars/main.js
```diff
@@ -55,27 +60,70 @@ function ghRepoGraphQl({ owner, repo }) {
 	return `${identifier}: repository(name: "${repo}", owner: "${owner}"){nameWithOwner stargazerCount}`;
 }
 
+/**
+ * Divide an array into multiple smaller array
+ *
+ * @param {Array} input
+ * @param {number} size
+ *
+ * @return {Array<Array>}
```
```suggestion
 * Divide an array into multiple smaller array
 * @param {Array} input
 * @param {number} size
 * @return {Array<Array>}
```

.github/actions/update-stars/main.js
```diff
@@ -55,27 +60,70 @@ function ghRepoGraphQl({ owner, repo }) {
...
+	}
+	return final;
+}
+
+/**
+ * Get the number of stars for all GitHub repositories.
+ *
+ * The result is a Map, where the key the repo name; and the value the number of stars.
+ *
+ * @returns {Promise<Record<string, number>>}
```
```suggestion
 * Get the number of stars for all GitHub repositories.
 * The result is a Map where the key the repo name and the value is the number of stars.
 * @returns {Promise<Record<string, number>>}
```

.github/actions/update-stars/main.js
```diff
@@ -89,25 +137,44 @@ function gitlabRepoGraphQl(name) {
 	return `${identifier}: project(fullPath: "${name}"){starCount fullPath}`;
 }
 
+/**
+ * Get the number of stars for all Gitlab repositories.
+ *
+ * The result is a Map, where the key the repo name; and the value the number of stars.
+ *
+ * @returns {Promise<Record<string, number>>}
```
```suggestion
 * Get the number of stars for all Gitlab repositories.
 * The result is a Map where the key the repo name and the value is the number of stars.
 * @returns {Promise<Record<string, number>>}
```

MacFJA:
> @MacFJA stars have disappeared from a lot of projects on https://sveltesociety.dev/components. I think it happened when I merged this PR. Maybe you could take a look?

(from https://github.com/svelte-society/sveltesociety.dev/pull/217#issuecomment-1673576811)

2 issues here:
 - The GitHub API have a more strict timeout since last time I run the script
 - The update only occurs once a month (https://github.com/svelte-society/sveltesociety.dev/pull/217#discussion_r1257534410) and no init was provided (my bad)

.github/actions/update-stars/main.js
```diff
@@ -55,27 +59,66 @@ function ghRepoGraphQl({ owner, repo }) {
...
 			.map((line) => [line.nameWithOwner.toLowerCase(), line.stargazerCount])
+			.sort()
 	);
 }
 
 // Gitlab
 
+/**
+ * Get all GitLab repositories path (relative to GitLab root)
+ *
```
```suggestion
```

.github/actions/update-stars/main.js
```diff
@@ -89,25 +131,42 @@ function gitlabRepoGraphQl(name) {
...
-	let lines = await doGraphQlQuery(gitlabGraphQlUrl, body);
+	core.info('Found ' + repoData.length + ' repositories');
+	const pagedRepoData = chunk(repoData, 100);
+	const pageCount = pagedRepoData.length;
+	core.debug('Divide the repositories into ' + pageCount + ' pages (of 100 repositories)');
+	let lines = [];
+	for (let index = 0; index < pageCount; index++) {
+		const page = pagedRepoData[index];
+		core.debug('Running GraphQL for page ' + (index + 1) + '/' + pageCount);
+		let body =
```
```suggestion
		const body =
```

benmccann:
thank you!!

benmccann:
Thanks for this @MacFJA!

We're getting closer, but I noticed a couple issues remaining. One was that it doesn't handle any repositories that have been renamed yet (e.g. the username changed or the repo was transferred to a GitHub org). I went ahead and updated the `.json` files manually to take care of those for now. However, there's still one issue left, which is that it's not working for any that have a URL that's not the GitHub URL. E.g. if you load https://sveltesociety.dev/components then the first one showing is `overmind-svelte` with a URL of https://overmindjs.org/. I think it's nice that we can link to homepages, but maybe we should change the existing `url` field to something like `repository` and then add a new optional `url` field where we can put any homepages that are different from the source code URL.

MacFJA:
Yeah, In the past I was testing how to get data.

One of my idea was to also retrieve information from NPM (if the package a well-defined, we have: name, description, repository, homepage, keywords, etc.). It could help to fill missing data.

----

What we can also do is, instead of replacing with the file only, we can replace the star count if found in the file, or leave as-is from the `components.json` (or tools, or templates)
(change this line: https://github.com/svelte-society/sveltesociety.dev/blob/main/src/lib/utils/stars.ts#L40 to `data.forEach((item) => (item.stars = getStarsCount(item.url) ?? item.stars));`)

benmccann:
There's not that many, so I don't mind filling in the missing GitHub URLs manually if we make the change I described above.

MacFJA:
Handled here: #438

PR #438, opened 2023-08-15
Add repository node to json

MacFJA:
- Add `repository` information to components, tools, templates data
- Add `repository` node to json files when needed
- Remove `stars` node from json (as it's only read from `stars.json`)
- Add link to repository on component card
- Update `stars.json` 
- Update the submission page


benmccann:
This is great! Thank you so much!

Do you think we could move the github URLs to always be contained within the `repository` field? I.e. use the the `url` field only if an entry has both a site url and repository URL present. We could probably write a little Node script to transform it or maybe use `jq` (I don't have any experience with the latter)

MacFJA:
> Do you think we could move the github URLs to always be contained within the `repository` field? I.e. use the the `url` field only if an entry has both a site url and repository URL present.

So swap the behavior ?

- `repository` is the required field
- `url` is an (optional) "superset" to add a specificity
- the link of the component/tool/template card is `repository` unless `url` is defined

But:
 - what about package that don't have public source?
   - don't make any URL mandatory ?
     - What happen is a package don't have a repository nor an URL?

MacFJA:
> We could probably write a little Node script to transform it or maybe use `jq` (I don't have any experience with the latter)

It's a one time action to do ;
so it can be easily be done by
- `jq`
- a node script
- `sed`
- an IDE/TextEditor seach/replace

benmccann:
> So swap the behavior ?
repository is the required field
url is an (optional) "superset" to add a specificity
the link of the component/tool/template card is repository unless url is defined

Yep, that's what I was thinking

> what about package that don't have public source?

Do we have any? I suppose we could make neither URL mandatory in that case

> What happen is a package don't have a repository nor an URL?

We should at the very least require one of the two fields to be present

> an IDE/TextEditor seach/replace

Duh. That's obviously the easiest :laughing:

.github/actions/update-stars/main.js
```diff
@@ -33,9 +33,9 @@ function gatherUrls() {
 	let templates = JSON.parse(readFileSync('src/routes/templates/templates.json'));
 
 	return [
-		...components.map((component) => component.url),
-		...tools.map((tool) => tool.url),
-		...templates.map((template) => template.url)
+		...components.map((component) => component.repository ?? component.url),
+		...tools.map((tool) => tool.repository ?? tool.url),
+		...templates.map((template) => template.repository ?? template.url)
```
```suggestion
		...components.map((component) => component.repository),
		...tools.map((tool) => tool.repository),
		...templates.map((template) => template.repository)
```

benmccann:
Amazing! Thank you!! This along with the autoupdate script is such a huge improvement!

PR #441, opened 2023-08-16
Add date on recipes

### LINKED ISSUE #420: Add publish date to articles.
Let's people know code is outdated.

MacFJA:
Close #420

- Add publication date on all recipes
- Add update date when the content is actually changed (exclude reformatting)
- Publish and update date are based on date from `git log` [^1]

[^1]: The exact command is: `git log --follow --format=%ad --date=iso-strict -- %PATH_TO_THE_FILE%`

### DIFF
--- a/src/lib/layouts/Recipe.svelte
+++ b/src/lib/layouts/Recipe.svelte
@@ -4,7 +4,6 @@
 	import { categories } from '$lib/stores/recipes';
 	import { page } from '$app/stores';
 	import Seo from '$lib/components/Seo.svelte';
-	import { format, formatDistanceToNow } from 'date-fns';
 
 	export let title;
 	export let published;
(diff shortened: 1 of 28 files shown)


src/lib/layouts/Recipe.svelte
```diff
@@ -4,8 +4,11 @@
 	import { categories } from '$lib/stores/recipes';
 	import { page } from '$app/stores';
 	import Seo from '$lib/components/Seo.svelte';
+	import { format, formatDistanceToNow } from 'date-fns';
```
benmccann: perhaps we could use `Intl.DateTimeFormat` rather than adding an additional dependency?

MacFJA: It's possible.
In fact, it's what I first wrote :sweat_smile: 

Let me change this

carlosvigil:
👏🏼

PR #440, opened 2023-08-16
Add CaptainCodeman components / packages

CaptainCodeman:

PR #442, opened 2023-08-16
remove listings with less than 15 stars

benmccann:
This will cut down on incoming PRs a lot to make them a lot easier to review and keep at least a bit of a floor on quality. There's a number of entries that got submitted a few years ago when the author initially created the library and it's never been touched or really even used since then. I ran the idea by Kev already and he's good with it.


benmccann:
Going to go ahead and close this for now after discussion on https://github.com/svelte-society/sveltesociety.dev/pull/440

PR #449, opened 2023-08-31
Add Prismic packages

samlfair:
Slice Machine didn't fit into any clear category on the tool pages so I marked it as "Content Management".


src/routes/tools/tools.json
```diff
@@ -282,5 +282,13 @@
 		"description": "Delegate events with on:* 🎉",
 		"npm": "svelte-preprocess-delegate-events",
 		"category": "Preprocessors"
+	},
+	{
+		"title": "@slicemachine/adapter-sveltekit",
+		"url": "https://prismic.io/docs/svelte",
+		"repository": "https://github.com/prismicio/slice-machine/tree/master/packages/adapter-sveltekit",
+		"description": "Turn your SvelteKit project into a website builder with Prismic's Slice Machine",
```
benmccann: the sveltekit adapters live in `components.json`. there's a category you can set for them. https://sveltesociety.dev/components#adapters

samlfair: These aren't SvelteKit adapters. It's a little confusing 😅 Slice Machine also has an adapter system. This is an adapter for Slice Machine, to make it compatible with SvelteKit. It has nothing to do with SvelteKit's adapter system. @benmccann

benmccann: Ok. I'd still add it to `components.json`, but use the `Integrations` category. I'd rather not add a new category on the tools page.

PR #457, opened 2023-09-15
Update components.json - add super-sitemap

jasongitmail:


src/routes/components/components.json
```diff
@@ -1,4 +1,19 @@
...
+		"description": "Automatic SvelteKit sitemap that makes it impossible to forget to add your paths.",
+		"npm": "sk-sitemap",
+		"category": "Developer Experience",
+		"tags": [
+			"components and libraries",
+			"integrations",
+			"layout and structure",
+			"typescript",
+			"sitemap",
+			"xml"
```
MacFJA: Those are new tags.

If there are no other components that can have it (and if it's the case, they need to be updated) I don't see why a tag that offer no alternatives should be added.

_(The list of tag is already long, I don't know if many people go to the struggle to search in it. Adding more tags without a real benefit don't seem like a good idea to reduce the issue, plus `sitemap` will be match be the search field anyway. But I can be wrong)_

jasongitmail: I can remove them. But would consider `sitemap` to be useful. The directory has similar `video` and  `images` tags.

Lmk what you think, will remove `xml` for now.

jasongitmail: I pushed a commit to update to package's new name on npm: `super-sitemap`.

I also removed the `sitemap` tag too, in addition to `xml` tag which was removed earlier. I still feel it's probably more useful for the community to keep the `sitemap` tag, but want to help make it easy to merge.

jasongitmail: @MacFJA ping on this

jasongitmail: Hi @MacFJA Just following up on this, seeing if we can get it merged please.

MacFJA: Yeah, of course.
Sorry, I was off a couple of days and completely forgot about it.

I see that the CI is failing, can you run a `pnpm run format` to fix it ? I will merge it as soon as the CI is OK

jasongitmail: Thanks! Formatted with Prettier. Passes now

PR #481, opened 2023-11-07
Fix SEO URL

lachlancollins:
The `Seo.svelte` file currently gets the host URL from an invalid `$page.host`, which should be `$page.url.host`.

The current live shows `<meta name="og:url" content="https://undefined/">`


MacFJA:
Thanks!

PR #483, opened 2023-11-12
Trial self-formatting PR workflow

lachlancollins:
Using `stefanzweifel/git-auto-commit-action` to auto-commit formatted file changes


lachlancollins:
Seems like it has issues [committing to forks](https://github.com/stefanzweifel/git-auto-commit-action/issues/278) - it would be possible to add as a workflow that runs on all code merged into main, but I feel like it's nicer to catch these styling issues during code review.

PR #484, opened 2023-11-12
Stricter npm field validation

lachlancollins:
Changes:

- Makes npm field required and uses stricter regex ([package-name-regex](https://www.npmjs.com/package/package-name-regex))
- Fixed properties for 5 entries in components.json 
- Moved 1 entry from components.json to templates.json
- Removed 1 entry from components.json that does not have a published npm package


src/lib/schemas.js
```diff
@@ -3,12 +3,12 @@ import { z } from 'zod';
...
-			.string()
-			.regex(/(@[\w-]+\/)?[\w-]+/)
-			.optional(),
+		npm: z.string().regex(/^(@[a-z0-9-]*\/)?[a-z0-9-]*$/),
 		url: z.string().url().optional(),
-		repository: z.string().url(),
+		repository: z
+			.string()
+			.url()
+			.refine((val) => val.includes('github.com') || val.includes('gitlab.com')),
```
MacFJA: I don't want to restrict repositories to only GitLab or GitHub.

Someone could want to host the code on Bitbucket, Sourceforge, etc.
We don't want to prevent them to add the component.

Today we don't have those cases, and so the script that update stars don't handle repositories others than GitHub or GitLab, but if there are others repository we will update the script

src/lib/schemas.js
```diff
@@ -3,12 +3,12 @@ import { z } from 'zod';
 export const componentsSchema = z.array(
 	z.object({
 		title: z.string(),
-		npm: z
-			.string()
-			.regex(/(@[\w-]+\/)?[\w-]+/)
-			.optional(),
+		npm: z.string().regex(/^(@[a-z0-9-]*\/)?[a-z0-9-]*$/),
```
MacFJA: > The `"name"` field contains your package's name, and must be lowercase and one word, and may contain hyphens and underscores.

Source: https://docs.npmjs.com/creating-a-package-json-file

underscore (`_`) is missing in your regex

src/routes/components/components.json
```diff
@@ -1002,13 +1003,6 @@
 		"title": "layercake",
 		"repository": "https://github.com/mhkeller/layercake"
 	},
-	{
-		"category": "CSS and Layout",
-		"description": "resizable layout component for svelte",
-		"tags": ["components and libraries", "layout and structure"],
-		"title": "svelte-layout-resizable",
-		"repository": "https://github.com/milahu/svelte-layout-resizable"
-	},
```
MacFJA, line 1011: Why removing this package ?

src/routes/components/components.json
```diff
@@ -1161,13 +1155,6 @@
 		"title": "storez",
 		"repository": "https://github.com/plrenaudin/svelte-storez"
 	},
-	{
-		"category": "Stores",
-		"description": "Template for client side svelte store",
-		"tags": ["components and libraries", "stores and state"],
-		"title": "SvelteStore",
-		"repository": "https://github.com/gitbreaker222/SvelteStore"
-	},
```
MacFJA, line 1170: Why removing this packages ?

src/lib/schemas.js
lachlancollins: Good point!

src/lib/schemas.js
lachlancollins: I've added [package-name-regex](https://www.npmjs.com/package/package-name-regex?activeTab=code) which should cover all cases 🤞

src/routes/components/components.json
lachlancollins: I couldn't find the package on npm - please let me know if you can find it!

src/routes/components/components.json
lachlancollins: I couldn't find the package on npm - please let me know if you can find it!

src/routes/components/components.json
MacFJA: The package have been deprecated, but can still be installed with:
```shell
npm install https://github.com/milahu/svelte-layout-resizable
# or
npm install milahu/svelte-layout-resizable
```
(ref: https://docs.npmjs.com/cli/v10/commands/npm-install)

But, do we want to allow that ? :thinking:

src/routes/components/components.json
MacFJA: Hmmm... after further analyzing/reading, it looks like it should be in `templates.json` more than in `components.json` :thinking:

src/routes/components/components.json
lachlancollins: The package itself can't be viewed if you visit https://npmjs.com/package/@milahu/svelte-layout-resizable since it's a private package, and https://npmjs.com/package/milahu/svelte-layout-resizable is just a 404 - since it's been deprecated anyway, probably not?

src/routes/components/components.json
lachlancollins: I've added SvelteStore to `templates.json` :)

lachlancollins:
Hi @MacFJA , just checking if you wanted any further changes? Thanks!

PR #490, opened 2023-11-14
fix: Remove explicit SvelteKit typing

lachlancollins:
As per https://svelte.dev/blog/zero-config-type-safety and discussion on #427

Also `export async function load(): Promise<PageLoad>` actually causes a TS error which is fixed by removing the type. I think this has to do with the difference between typing function declarations and function expressions; the [SvelteKit load docs](https://kit.svelte.dev/docs/load) show how to type function expressions which do work in this case.

### DIFF
--- a/src/routes/events/+page.svelte
+++ b/src/routes/events/+page.svelte
@@ -2,8 +2,7 @@
 	import Societies from '$lib/components/Societies/index.svelte';
 	import EventListElement from '$lib/components/EventListElement/index.svelte';
 	import Seo from '$lib/components/Seo.svelte';
-	import type { PageData } from './$types';
-	export let data: PageData;
+	export let data;
 </script>
 
 <Seo title="Events" />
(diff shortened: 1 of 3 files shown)


src/routes/events/+page.svelte
```diff
@@ -2,8 +2,7 @@
 	import Societies from '$lib/components/Societies/index.svelte';
 	import EventListElement from '$lib/components/EventListElement/index.svelte';
 	import Seo from '$lib/components/Seo.svelte';
-	import type { PageData } from './$types';
-	export let data: PageData;
```
MacFJA, line 6: If `data` was used inside the `<script lang="ts">` I would argue that removing the type is only applicable if you use the Svelte(Kit) Language Server, and be extra explicit won't harm anyone.

But, as `data` is only used inside the markup, where only the Svelte Language server can suggest completion on variable inside `{}` and `{# ...}` and `{@ ...}` I won't be overly preventive

MacFJA (approved):

PR #491, opened 2023-11-14
fix: Update type for getPages function

lachlancollins:
Inferred type before:

```
let event: {
    [key: string]: string;
    filename: string;
    path: string;
    title?: string;
}
```

Inferred type after:

```
let event: {
    filename: string;
    path: string;
    title: string;
    layout: string;
    date: string;
}
```


src/routes/pageList.ts
```diff
@@ -1,6 +1,9 @@
 export async function getPages(
-	metaGlob: Array<string>
-): Promise<Array<{ [key: string]: string; filename: string; path: string; title?: string }>> {
```
MacFJA, line 3: The `[key: string]: string` is still needed for building the hierarchy of recipes (a `children` key is injected later)

MacFJA:
I think we can go a bit further on this.
I will open a PR on your PR (PR-ception :sweat_smile: ) with what I have in mind (it's here: https://github.com/lachlancollins/sveltesociety.dev/pull/1)

PR #494, opened 2023-11-15
Add wstaeblein repos

wstaeblein:


MacFJA:
Your PR have formatting issue.

Can you run `pnpm run format` to solve them?

----

Your PR is also in conflict as another PR was merged before yours, so if you can rebase it, it would be great.

lachlancollins:
Hi @wstaeblein , thanks for your submissions! Unfortunately, the `components.json` file requires packages to be published on NPM, since we obtain metadata about the package from here. I understand these are simple components, however it also isn't straight-forward for a user to find the right code they need to copy into their own project.

If you would like to publish your package, you can use [this template](https://github.com/lachlancollins/svelte-library-template) I created as an example structure. Let me know if I can do anything to help!

PR #499, opened 2023-11-22
feat: Add script to get npm package data

lachlancollins:
## 🎯 Changes

This PR adds auto-fetching package data from npm. The `scripts/updateNpm.js` script runs the npm CLI in parallel to get info listed [here](https://docs.npmjs.com/cli/v7/commands/npm-view). For now, I've set it to save the latest release date, latest version (useful to show how up-to-date the info is and for other tools such as publint), and svelte peerDep (if available).

For now, the latest version and date are added to the package card (if it exists). The date is rendered as a relative date using dayjs. I've moved the repo button from the bottom-left to the top-right of the card (this was the most space-efficient arrangement of stars, repo and date I could pick).

### Before:

image

### After:

image

## ✅ Checklist


scripts/updateNpm.js
```diff
@@ -0,0 +1,47 @@
...
+const data = componentsSchema.parse(components);
+
+const { versions, dates, support } = await Promise.all(
+	data.map(async (pkg) => {
+		try {
+			return await processPackage(pkg);
+		} catch (error) {
+			console.log(error.message);
+		}
+	})
```
MacFJA: Unnecessary `async`/`await`
```suggestion
	data.map((pkg) => processPackage(pkg)
		.catch (error => console.log(error.message))
	})
```

scripts/updateNpm.js
```diff
@@ -0,0 +1,47 @@
...
+	let support = {};
+	for (const value of values) {
+		if (value) {
+			versions[value.name] = value.version;
+			dates[value.name] = value.date;
+			support[value.name] = value.support;
+		}
+	}
+	return { versions, dates, support };
+});
```
MacFJA: This doesn't add anything, and it's probably worse in terms of performance and memory usage.

It can be easier to read of some, but your version is pretty logical (a `for` loop, a `if` and 3 variables is way more standard than a `.filter(Boolean)` and a `.reduce`)

My version duplicate the data 2 times in memory (one in the `.filter`, one in the `.reduce`) where you version only once (result of the `for` loop)  
_My version won't probably never by optimized by any compiler_

```suggestion
).then((values) => values
	.filter(Boolean)
	.reduce((result, value) => {
		result.versions[value.name] = value.version;
		result.dates[value.name] = value.date;
		result.support[value.name] = value.support;
		return result;
	}, {versions: {}, dates: {}, support: {}})
);
```

MacFJA (commented):
I'm not sure about creating a file by data type :thinking: 

If the data were to be updated separately, it could make sens, but here everything is updated together at each run

lachlancollins:
> I'm not sure about creating a file by data type 🤔
> 
> If the data were to be updated separately, it could make sens, but here everything is updated together at each run

Yep for sure, definitely makes sense - I started off just with a `versions.json` then it grew!

Which one of these do you think would be better?

1.
```
"versions": {
		"super-sitemap": "0.14.11",
		"svelte-scrollactive": "0.0.9",
		...
```

2.
```
"super-sitemap": {
		"version": "0.14.11",
		"date": "2023-11-18T20:28:44.230Z",
		...
```

src/lib/utils/injectNpmData.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const injectNpmData = (input: z.infer<typeof componentsSchema>) => {
+	const output = [];
+	for (const item of input) {
+		const extra = npm[item.npm];
+		if (extra) {
+			output.push({ ...item, ...extra });
+		} else {
+			output.push(item);
+		}
```
MacFJA:
```suggestion
		const extra = npm[item.npm] ?? {};
		output.push({ ...item, ...extra });
```

MacFJA (commented):
I would reduce the font size of the date.

It's a bit too "in your face" right now. It has the same font-size as the description, making it as important as the description.

Also, I don't find use `<code>` revelant here, it's a "none-technical" information nor a number

PR #500, opened 2023-11-23
Update to typescript eslint v6

lachlancollins:
Changes:

- Bumped @typescript-eslint/eslint-plugin and @typescript-eslint/parser to v6.12.0
- Updated other eslint plugins
- Updated `.eslintrc.cjs` as described on the [v5 to v6 blog](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/)
- Disabled two new rules included by `plugin:@typescript-eslint/stylistic` that weren't enabled previously and were now failing
- Added `@ts-check` comment to get type warnings in IDE
- Added `.eslintignore` (necessary because when TS eslint tries to run type rules against files not included in `tsconfig.json`, it causes an error - this included the scripts files, svelte.config.js and vite.config.js)


.eslintignore
```diff
@@ -0,0 +1,6 @@
+/.svelte-kit
+/build
+/node_modules
+/scripts
+svelte.config.js
+vite.config.js
```
MacFJA: Should we ignore code issue in those files ?

For me, they still need to be valid, even if they are configuration or "not production" files

lachlancollins: Yeah it's annoying, I described the reason for adding this above - since the tsconfig `included` array can't be extended in sveltekit as it would override the auto-generated config, TS-eslint throws an error on these files because they aren't included in TS, but it is attempting to lint them with type context. I'll see if there is a workaround.

lachlancollins: Turns out there is a way (https://kit.svelte.dev/docs/configuration#typescript) - I had no idea this existed! Will commit soon.

lachlancollins: Solved!

PR #504, opened 2023-11-28
chore: Set CI to use Node v20

lachlancollins:
Changes:

- Updated from Node 18 (most actions) and Node 16 (update-stars action only) to Node 20
- Pull node version from `.nvmrc` (sets GH and Netlify version simultaneously)
- Updated external actions to current versions
- Removed unnecessary node registry-url (not using private repo)


lachlancollins:
Hi @MacFJA , could you please check out this PR?

PR #514, opened 2023-12-17
feat: Use `@sindresorhus/slugify` to generate category IDs

lachlancollins:
## 🎯 Changes

`SearchableJson.svelte` currently requires each category to be manually mapped to a slugified category ID. This package handles the conversion and will be more flexible if new categories are added. Since `SearchableJson.svelte` is shared, I've moved it from `/src/routes` into `/src/lib`.

## ✅ Checklist


MacFJA:
The only issue I have it's we will lose any previously used anchor.

If someone post a list to a category before, the link will not scroll to the category (but will stay at the top of the page)

But I think it's a minor drawback

PR #517, opened 2023-12-17
feat: Use netlify adapter with edge functions

lachlancollins:
## 🎯 Changes

Use Netlify edge functions rather than static adapter: https://kit.svelte.dev/docs/adapter-netlify#netlify-edge-functions

## ✅ Checklist


MacFJA:
What are the advantages of using Netlify Edge over the Static generation ?

lachlancollins:
> What are the advantages of using Netlify Edge over the Static generation ?

For this PR, there should be no change at all from adapter-static to adapter-netlify, since all pages are still prerendered and therefore rapidly served as html via the same netlify CDN. This is what I've observed from the deploy preview.

However, with a non-static adapter, it would be possible to add server-side logic in the future (e.g. could process large JSON files server-side and only send what is necessary to the browser). Effectively, adapter-netlify seems to be a superset of  adapter-static with the current setup.

PR #533, opened 2023-12-20
feat: Submitted data restructure

### LINKED ISSUE #532: Proposal: Submitted data restructure
## 💭 Problem

I've found that the "Components", "Tools", and "Templates" distinction isn't always logical. For example, `@sveltejs/adapter-auto` is in `components.json`, while `vite-plugin-svelte` is in `tools.json`. "Components" in particular is _far_ more than just UI packages. It would be easy to miss a package because you don't know where to look.

In comparison, I've found the SolidJS Ecosystem pages to be very well structured and logical. They divide their content into ["Packages"](https://www.solidjs.com/ecosystem) and ["Resources"](https://www.solidjs.com/resources).

## 🚀 Proposal

I think the packages vs resources division makes a lot of sense. Here is how I think the current data could fit this:

### `/packages`
- Things that you can `npm install` and add to `package.json`.
- Sourced from current `components.json` and `tools.json`.
- Using the above example, both `@sveltejs/adapter-auto` and `vite-plugin-svelte` would be included on this page.

### `/resources`

- Templates, sample projects, extensions, tutorials, podcasts, etc.
- Sourced from current `templates.json`, `tools.json`, and `/resources`.
- For example, `SvelteKit i18n Template`, `Svelte DevTools` and `Svelte for VS Code` would be included on this page.
- Opens the door for more resource submissions (e.g. I can't find any references to Joy of Code's channel)

lachlancollins:
## 🎯 Changes

Closes #532

See discussion on #532. Merges `/tools` contents into other existing locations. Renames `/components` to `/packages`. Redirects old pages to new page. Turns books and videos on `/resources` into JSON data for simpler additions.

## ✅ Checklist


src/routes/resources/+page.svelte
```diff
@@ -1,96 +1,51 @@
...
 	<h2>Books</h2>
 
-	<div>There are a few books from major publishers:</div>
 	<ul>
-		{#each booksFromPublisher as { name, link, author }}
-			<li><a href={link} target="_blank">{name}</a> by {author}</li>
-		{/each}
-	</ul>
-
-	<div>As well as a couple self-published books:</div>
```
MacFJA: We lose the indication of self-published, which is a different involvement and process

src/routes/resources/+page.svelte
```diff
@@ -1,96 +1,51 @@
...
-		<li><a href="https://egghead.io/browse/frameworks/svelte" target="_blank">Egghead</a></li>
-		<li>
-			<a href="https://www.udemy.com/courses/search/?q=sveltejs+svelte" target="_blank">Udemy</a> (Note:
-			Udemy frequently has discounts over 90%)
-		</li>
-		<li><a href="https://www.pluralsight.com/search?q=svelte" target="_blank">Pluralsight</a></li>
-	</ul>
+	<h2>Extensions</h2>
 
-	<div>Finally, there are also YouTube channels and playlists that teach Svelte:</div>
```
MacFJA: We lose the distinction between learning platform and YouTube.

Learning platform often have extra bits (documents, projects, support, etc.)

src/routes/components/+page.server.ts
```diff
@@ -0,0 +1,5 @@
+import { redirect } from '@sveltejs/kit';
+
+export const load = async () => {
+	redirect(302, '/packages');
```
benmccann:
```suggestion
	redirect(308, '/packages');
```

src/routes/tools/+page.server.ts
```diff
@@ -0,0 +1,5 @@
+import { redirect } from '@sveltejs/kit';
+
+export const load = async () => {
+	redirect(302, '/packages');
```
benmccann:
```suggestion
	redirect(308, '/packages');
```

src/routes/resources/+page.svelte
lachlancollins: Good point, fixed!

src/routes/resources/+page.svelte
lachlancollins: Fixed!

Issue #443, opened 2023-08-18
Fetch more data from GitHub


benmccann:
@CaptainCodeman suggested it's too difficult to list a new component at the moment. It made me think about how we could simplify things and I do think things could be quite a bit easier. A lot of the info we need is already available on GitHub, so we should just fetch it from there rather than making the user enter it all.

We could make it so that the user only has to enter the title, npm package name, GitHub URL, category, and tags

Once we have the repo URL we can then fetch the homepage URL, description, and star count from GitHub. @MacFJA just added a script to fetch the star count, so hopefully it wouldn't be too hard to extend. We could also fetch the last commit date and whether the repo is archived from GitHub, which would help a lot in identifying stale repositories.

I think we should get rid of the `addedOn` field. If you have a library that's 5 years old, who cares if it was added to sveltesociety.dev 3 years ago or last week?

CaptainCodeman:
I had a play around last evening doing exactly this and it works well - all you need to enter though is the `npm` package name and everything else can be retrieved from that.

This is a list of components (in no particular order) rendered doing that, with images coming from the open-graph / twitter cards taken from the project homepages where available:

<img width="1668" alt="Screenshot 2023-08-18 at 10 25 02 AM" src="https://github.com/svelte-society/sveltesociety.dev/assets/304910/f45c4222-e457-40ea-b5da-48e8f7f89a97">

The readme can be rendered gfm style on a detail page and the original creation date and last version published date both available from npm (which are probably more useful than the date listed on the site). I'd suggest some vector similarity indexing to group related components, and / or use a convention of keywords to decide what categories things go into, and would show similar / other component cards based on what is being viewed (this was storing things in Redis, which can provide the vector similarity indexing):

<img width="1633" alt="Screenshot 2023-08-18 at 10 36 24 AM" src="https://github.com/svelte-society/sveltesociety.dev/assets/304910/b2c3f32f-d01f-4879-af8a-c91564659061">

You would probably want to have submissions require authentication, and an approval process before they showed up to protect against abuse, but those would both be fairly simple and you could then auto-approve based on previous approvals for the same account to cut down on work (maybe also just allow people to view verified / un-verified entries as a toggle).

benmccann:
Wow, that looks great!

[...]

PR #541, opened 2023-12-21
feat: Merge categories/tags and improve filtering

lachlancollins:
## 💭 Problem

This was prompted in part by discussion on #443. Tags are effectively unlimited and unregulated right now; however, categories are also quite limited since each entry can only have one. Filtering based on category/tags via the select fields is an okay experience - but it could be better!

## 🎯 Changes

Heavily inspired by [bestofjs.org](https://bestofjs.org/tags). Each item will be allowed multiple, pre-defined categories (or tags, the name doesn't matter). This would be a merge of current categories and useful tags to keep. By default, all entries are displayed. A collection of buttons let you choose which categories/tags you want displayed. When you select a category/tag, the list of available categories/tags narrows based on your current selection. This utilises a server load function, and the URL is shareable due to search params.

### Before

image

### After

image

image

image

## ✅ Checklist


benmccann:
Honestly, I'm not a big fan of this as implemented. We commonly do things like link to https://sveltesociety.dev/packages#svelte-kit-adapters. There's no way to do that now as everything's just mixed together and that filter doesn't exist. It's also going to break tons of links that we have across svelte.dev and kit.svelte.dev. We could clean those up, but we'd need ways to keep the existing functionality for that to be possible

kevmodrome:
This looks great!

lachlancollins:
> Honestly, I'm not a big fan of this as implemented. We commonly do things like link to [sveltesociety.dev/packages#svelte-kit-adapters](https://sveltesociety.dev/packages#svelte-kit-adapters). There's no way to do that now as everything's just mixed together and that filter doesn't exist. It's also going to break tons of links that we have across svelte.dev and kit.svelte.dev. We could clean those up, but we'd need ways to keep the existing functionality for that to be possible

@benmccann oh wow fair enough, I had no idea it was referenced extensively in the svelte/kit docs. For that above link, with this change it would become https://sveltesociety.dev/packages?tag=svelte-kit-adapters. I guess the worst case scenario for old links is that users get taken to the top of the page rather than the category ID that was linked. I would argue this new solution is more robust than since it only loads the contents of the specific category/tag in the link. Either way, I'm not too fussed if this proposal isn't feasible since it was fun to implement!

kevmodrome:
There is a feature in Cloudflare to do bulk redirects - maybe that's something we can explore?

kevmodrome:
Given that we get the same functionality as before - would it be fine to merge this if we updated all the links on the Svelte site? @benmccann

benmccann:
Let me give this another look to verify whether things like the `?tag=svelte-kit-adapters` filter will let us reproduce what we were depending on previously. I remember there being other issues we might need to address, but I can't check at the moment because the Netlify deploy is broken. Do we need to close this PR and open a new one to see the Vercel deploy preview or was the Netlify one taken down and the Vercel one not added yet?

kevmodrome:
I think so.

kevmodrome:
Testing Cloudflare Preview

benmccann:
When I visit https://sveltesociety-dev-preview-git-fork-lachlan-97a95c-sveltesociety.vercel.app/packages it gives me a 500 error

lachlancollins:
> When I visit [sveltesociety-dev-preview-git-fork-lachlan-97a95c-sveltesociety.vercel.app/packages](https://sveltesociety-dev-preview-git-fork-lachlan-97a95c-sveltesociety.vercel.app/packages) it gives me a 500 error

@benmccann this is fixed by #546 (the site wasn't using adapter-vercel)

benmccann:
> For that above link, with this change it would become https://sveltesociety.dev/packages?tag=svelte-kit-adapters

Doesn't seem to work for me. This returns no results: https://sveltesociety-dev-preview-git-fork-lachlan-97a95c-sveltesociety.vercel.app/packages?tag=svelte-kit-adapters

lachlancollins:
> > For that above link, with this change it would become https://sveltesociety.dev/packages?tag=svelte-kit-adapters
> 
> Doesn't seem to work for me. This returns no results: https://sveltesociety-dev-preview-git-fork-lachlan-97a95c-sveltesociety.vercel.app/packages?tag=svelte-kit-adapters

Yes sorry, I haven't actually merged categories and tags yet, I wanted to get some feedback before I put that effort in! Only existing tags (not categories) can be selected right now. For example: https://sveltesociety-dev-preview-git-fork-lachlan-97a95c-sveltesociety.vercel.app/packages?tag=notifications

lachlancollins:
Hi @benmccann , would you be able to give this another look? Thanks!

PR #629, opened 2024-06-14
added preprocessor "svelte-remove-attributes"

jacobbogers:
preprocessor to remove attributes (like `data-testid` from production builds)

## 🎯 Change

added entry for a preprocessor

## ✅ Checklist


lachlancollins (changes requested):
Hi @jacobbogers , you need to fix formatting, and the npm field should be the package name rather than a URL.

jacobbogers:
> Hi @jacobbogers , you need to fix formatting, and the npm field should be the package name rather than a URL.

I have corrected, it is not an url but actual package name

lachlancollins (changes requested):
Still failing linting, you need to fix the prettier issues.

jacobbogers:
pnpm does not work for me, I am running prettier directly with --write, it did produce changes in package.json, so hope this works

jacobbogers:
image

jacobbogers:
`"lint": "prettier --check **/* --ignore-unknown && eslint . && node scripts/validateData.js",`

I am on windows, (11) remove the `'` around the  `**/*` and it will work for me,  does it work on linux/mac without `'` ?

lachlancollins (approved):

PR #667, opened 2024-08-15
Refactoring admin

MacFJA:
Refactoring of the table in the admin

Add "magic" action on Input component to allow auto-fill (ex: generating the slug from the title)


kevmodrome (approved):
Thanks! I had plans to refactor things a bit. Looks good!

PR #760, opened 2025-04-29
feat: tweak saved page link button ux in sidebar

ghostdevv:
## 🎯 Changes

I was clicking on the saved button in the sidebar, and nothing was happening so I was confused. This should clear that up by redirecting my to auth if I'm not logged in.

Alternatives considered:
- Grey out the button if you're not logged in
- Allow it to be clicked but toast the user if they're not logged in, that they should be

## ✅ Checklist

- [ ] I have run `pnpm run lint` locally on my changes

### DIFF
--- a/src/routes/(app)/(public)/saved/+page.server.ts
+++ b/src/routes/(app)/(public)/saved/+page.server.ts
@@ -1,9 +1,8 @@
 import { error, redirect } from '@sveltejs/kit'
 import type { PageServerLoad } from './$types'
-import type { Content } from '$lib/types/content'
 
 export const load = (async ({ locals, url }) => {
-	if (!locals?.user?.id) redirect(302, '/')
+	if (!locals?.user?.id) redirect(302, '/auth/github')
 
 	try {
 		// Get pagination parameters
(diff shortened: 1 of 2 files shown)


MacFJA:
If you just redirect to the login page, you will lose what you were doing.

I think it's better to grey out the button and add a tooltip to indicate that you need to be logged in to use the feature

---

Or update the enhancing (https://github.com/svelte-society/sveltesociety.dev/blob/v2/src/lib/ui/ContentCard.svelte#L33) to handle the result and display a feedback on what happened (so, in the case you log in another tab, the button won't stay in a weird state until you refresh the page)

ghostdevv:
I tweaked it to grey out the button for simplicity, wdyt?

ghostdevv:
~~My primary goal of this was achieved in f15e28f17c4781f1767fb7613d57d28aff005acc, so closing this :pray:~~

ghostdevv:
Actually, I'll let @kevmodrome in case you want this behaviour of showing a tooltip rather than hiding the button

kevmodrome (approved):

PR #787, opened 2025-08-24
Added @thebrenny/svelte-adapter-h3

TheBrenny:
adding adapter-h3 to (legacy) packages.json

H3 hoists middleware up to the global middleware stack when mounting sub-apps. This isn't good because svelte effectively only uses middlewares. This was worked around by using `app.all("/**/*", (e) => svelteApp.fetch(e))`.


kevmodrome:
Please make a submission on the new website https://prod.sveltesociety.dev

PR #803, opened 2025-10-08
feat: refresh og images every 14 days

karimfromjordan:
> [!CAUTION]
> Let's first deploy this to `staging` and test it. After that I'll check and make sure everything is fine. Then we can deploy it to `production`.

This extends the `/files` endpoint and refreshes an image

- if it is older than `14 days` 
- if it currently does not exist on disc and needs to "heal" because the download previously failed (rate limit)


kevmodrome (approved):
LGTM

PR #812, opened 2025-10-10
fix: rename thumbnail url in database

karimfromjordan:
This won't fix the thumbnails on `staging` because the database there is outdated but it will make sure that everything works in `prod`.


kevmodrome:
LGTM, we probably want to periodically fetch a prod database to put into staging anyway

PR #826, opened 2025-10-22
fix account dropdown obscured

PatrickG:
Follow up to #825
Fixes account dropdown being obscured by the header.


kevmodrome:
LGTM - TY!

PR #907, opened 2025-12-10
feat: discord logger for `500` errors

karimfromjordan:
We need to set the `DISCORD_WEBHOOK` environment variable when deploying.


kevmodrome (approved):
LGTM!

I think the only thing we need to do is to add an empty variable to the .env.example file

